PR #81, opened 2015-06-16
Run scalatest on Travis CI

nlochschmidt:
So here is my proposal for getting all the tests to execute on Travis CI (see #2) by defining a sbt multi-project that aggregates all the other projects. 

It is maybe not the most elegant way, but I couldn't figure out how to dynamically define all problems as subprojects in the `build.sbt` and therefore used the *old* way of defining the build in `Build.scala`. An alternative would be to let Travis write a `build.sbt` on the fly.

A little script is also removing all the pending statements before tests are being executed. 

The only issue I found was in the `BankAccount` problem, where the timing sensitive test failed because of a timeout. The issue here is most likely that the Scalatest Testrunner executes the tests in all project in parallel which means individual tests run slower compared to being run in isolation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
-language: bash
+language: scala
+scala:
+  - 2.10.3
+
+before_script:
+  - bin/fetch-configlet
+  - sbt test:compile
 
 script:
-  - bin/fetch-configlet
   - bin/configlet .
+  - sbt test
--- a/bank-account/src/test/scala/BankAccountTest.scala
+++ b/bank-account/src/test/scala/BankAccountTest.scala
@@ -1,7 +1,7 @@
-import org.scalatest.concurrent.Conductors
+import org.scalatest.concurrent.{IntegrationPatience, Conductors}
 import org.scalatest.{Matchers, FunSuite}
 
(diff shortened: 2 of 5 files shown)


kytrinyx:
I don't know enough to discuss the options with build.sbt / Build.scala -- the one thing that comes to mind here, though, is that I'd be worried that if someone runs the build script locally, they might forget to put the `pending` statements back before committing. How about copying everything to a temporary directory before running the build?

nlochschmidt:
I removed the script that modified the test files in place and instead put some magic into the build definition so that the test sources are compiled without the `pending`s while leaving the original files intact.

kytrinyx:
Cool. If @ricemery doesn't have any suggestions, then I would say that this is ready to be squashed and merged.

ricemery:
I am ok with it.

kytrinyx:
Thanks @ricemery. @nlochschmidt Would you mind squashing the commits? If not, let me know and I can squash them locally before merging.

nlochschmidt:
@kytrinyx done

kytrinyx:
Thanks!

PR #83, opened 2015-08-16
Fix travis build and test against java 7 and 8

nlochschmidt:
Hi @ricemery 

I noticed that the build on Travis was broken. I made a mistake naming the build definition `ThisBuild` which shadows the sbt internal `ThisBuild` scope. Sorry about that.

In addition I think it makes sense to test both against java 7 and 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
+sudo: false
 language: scala
 scala:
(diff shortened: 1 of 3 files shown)


kytrinyx:
Thanks!

PR #84, opened 2015-09-13
Make nucleotide-count example and tests more scaly

astradamus:
- Made nucleotidesCount a lazy val instead of a def in example solution. The return result would never change, but was running through validation and counting every time it was called.
 - Replaced deprecated exception syntax in test suite.
 - Changed nucleotide terms in test suite from ribonucleoside form to nucleobase form (i.e. adenosine to adenine) to be consistent with terms used in exercise readme.
 - Removed test suite's insistence that solution provide two different members that, in Scala, do the same thing (count vs. nucleotideCounts). This insistence, along with specifically requiring an InvalidArgumentException, railroads solutions towards too small a possible set. All calls to count are replaced with calls to nucleotideCounts, because Scala maps already provide the functionality of count (except that it throws a NoSuchElementException instead of an IAE), and the test now only expects that SOME RuntimeException be thrown, rather than an IAE specifically, allowing solvers to judge for themselves if the added specificity of a separate accessor that translates the map's NSEE to an IAE is worth the verbosity.

### DIFF
--- a/nucleotide-count/example.scala
+++ b/nucleotide-count/example.scala
@@ -1,26 +1,16 @@
 class DNA(strand: String) {
-  strand.foreach(validate)
+  strand foreach validate
 
-  def count(nucleotide: Char) = {
-    validate(nucleotide)
-    strand.count(_ == nucleotide)
+  lazy val nucleotideCounts = {
+    Map('A' -> 0, 'T' -> 0, 'C' -> 0, 'G' -> 0) ++
+    strand.toCharArray.groupBy(identity).mapValues(_.length)
   }
 
-  def nucleotideCounts = Map(
-    'A' -> count('A'),
-    'T' -> count('T'),
-    'C' -> count('C'),
-    'G' -> count('G')
-  )
-
   private def validate(nucleotide: Char) =
-    if (!isNucleotide(nucleotide))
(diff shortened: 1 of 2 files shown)


astradamus:
As another note, I couldn't seem to figure out what the point was of the test "should count a nucleotide only once." I looked at the Java version and it made equally little sense. Is it a weird artifact from when this test was translated from some other language? It doesn't actually seem to do anything that ensures counting is not repeated.

kytrinyx:
I think that's an artifact from the Ruby version, where it was added because some beginners were doing very, very strange things that passed the tests but were a horrible idea. I think it's fine to remove it. It should come up in the conversation, anyhow.

kytrinyx:
Thanks!

astradamus:
Happy to help! I have to admit I'm curious what those strange solutions were.

kytrinyx:
I don't recall off the top of my head, but I think that the method that should have been a straight read was mutating the original strings, so that if you read the value twice, it came out differently.

astradamus:
Holy hell Strings are mutable in Ruby! That's not a question mark because I Googled it. That sounds frightening.

kytrinyx:
Indeed :)

PR #87, opened 2015-12-09
Create ABOUT.md

dffuller:


kytrinyx:
Sweet! How about also having something short about what Scala is typically used for in the real world?

PR #95, opened 2015-12-28
Add forth exercise

ricemery:
Add forth exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,7 +64,8 @@
     "simple-cipher",
     "connect",
     "parallel-letter-frequency",
-    "acronym"
+    "acronym",
+    "forth"
   ],
   "deprecated": [
 
--- a/forth/build.sbt
+++ b/forth/build.sbt
@@ -1,0 +1,6 @@
+scalaVersion := "2.11.7"
+
+sourcesInBase := false
+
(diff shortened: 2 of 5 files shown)


ricemery:
@kytrinyx - question for you. The Travis CI build is failing on my pull request. The problem is that in the src/main/scala directory I have a .scala file that needs to be filled out by the user. The file contains a couple of class definitions that define the interface needed by the unit tests.

In the root is the example.scala file that has a "real" solution. The build.sbt is written to compile the file in src/main/scala, since this is what the user will want. The example.scala is ignored by build.sbt. 

Travis CI is trying to compile and run the unit tests, which will fail since the example.scala is ignored. 

Any suggestions on how to deal with this case?  I could add Ignore flags on all the unit tests. I'm not sure if there is a better way.

Thanks

kytrinyx:
Is there a a way to make the build ignore the stub file when running the tests against examlpe.scala?

ricemery:
@kytrinyx - I was making this harder than I needed to. I changed the stub to include traits instead. So, now the user will have a couple of traits to implement. And, the CI build should pass.

Thanks

kytrinyx:
That sounds good.

PR #100, opened 2016-01-26
Move exercises to subdirectory. Fixes #97

### LINKED ISSUE #97: Move exercises to subdirectory
The problems api (x-api) now supports having exercises collected in a subdirectory
named `exercises`.

That is to say that instead of having a mix of `bin`, `docs`, and individual exercises,
we can have `bin`, `docs`, and `exercises` in the root of the repository, and all
the exercises collected in a subdirectory.

In other words, instead of this:

```
x{TRACK_ID}/
├── LICENSE
├── README.md
├── bin
│   └── fetch-configlet
├── bowling
│   ├── bowling_test.ext
│   └── example.ext
├── clock
│   ├── clock_test.ext
│   └── example.ext
├── config.json
└── docs
│   ├── ABOUT.md
│   └── img
... etc
```

we can have something like this:

```
x{TRACK_ID}/
├── LICENSE
├── README.md
├── bin
│   └── fetch-configlet
├── config.json
├── docs
│   ├── ABOUT.md
│   └── img
├── exercises
│   ├── bowling
│   │   ├── bowling_test.ext
│   │   └── example.ext
│   └── clock
│       ├── clock_test.ext
│       └── example.ext
... etc
```

This has already been deployed to production, so it's safe to make this change whenever you have time.

ricemery:
Move exercises to subdirectory. Fixes #97


kytrinyx:
I don't understand why configlet is failing. It's passing locally. I wonder if there's some way it might be getting an old version of the tool.

kytrinyx:
I'm pushing a commit to see what version of configlet the build is downloading.

kytrinyx:
OK, it's getting the correct version of configlet. There must be some difference between mac os x and linux in how it's doing paths or something. I'll debug this today and fix configlet.

ricemery:
Maybe a case sensitivity issue. OS X's file system is case insensitive.

kytrinyx:
Good call, that could be it.

kytrinyx:
I found the problem, I'll push a fix and re-trigger the build.

Meanwhile (and not related to this)... there are lots of different styles of test names. Is this intentional?

- `SaddlePointsSpecs.scala`
- `scrabble_score_test.scala`
- `SecretHandshakeTest.scala`

ricemery:
No good reason on the different style test names. Scala doesn't force the filename to match the class definition. So.. The test names are just whatever made sense to the author at the time of creation.

kytrinyx:
OK, cool. We should probably normalize at some point :)

kytrinyx:
The build is fixed.

PR #123, opened 2016-06-27
Change regex from \w to [A-Z]

JPBetley:
The regex used in the test doesn't explicitly search for the format given in the problem documentation. This makes it explicit that the first two characters need to be an uppercase letter

### DIFF
--- a/exercises/robot-name/src/test/scala/robot_name_test.scala
+++ b/exercises/robot-name/src/test/scala/robot_name_test.scala
@@ -1,7 +1,7 @@
 import org.scalatest._
 
 class RobotNameSpecs extends FunSpec with Matchers {
-  val nameRegex = """\w{2}\d{3}"""
+  val nameRegex = """[A-Z]{2}\d{3}"""
 
   it ("has a name") {
     new Robot().name should fullyMatch regex (nameRegex)


ricemery:
Thanks. 
Question though. I don't see an explicit documentation that states that the first two characters must be uppercase. I see a couple of examples with uppercase characters. But, no explicit definition of the format. 
From the readme -

> The first time you boot them up, a random name is generated, such as RX837 or BC811.
> 
> Every once in a while we need to reset a robot to its factory settings, which means that their name gets wiped. The next time you ask, it will respond with a new random name.
> 
> The names must be random: they should not follow a predictable sequence. Random names means a risk of collisions. Your solution should not allow the use of the same name twice when avoidable. In some exercism language tracks there are tests to ensure that the same name is never used twice.

It seems like the unit test is providing the only concrete definition.

JPBetley:
True, and I think that's an issue with the documentation of the problem as well. I'm not 100% sure where that documentation resides, but if you could point it out to me, I'd be happy to make a PR to it.

The reason for this PR was because of a comment I received on my own solution to this problem, noting that the first two chars should be uppercase letters. I looked at several other languages test for "Robot Name" and saw that they explicitly check for two uppercase letters at the start as well. Here are a few examples:
- Rust - https://github.com/exercism/xrust/blob/master/exercises/robot-name/tests/robot-name.rs
- Java - https://github.com/exercism/xjava/blob/master/exercises/robot-name/src/test/java/RobotTest.java
- Haskell - https://github.com/exercism/xhaskell/blob/master/exercises/robot-name/robot-name_test.hs
- Python - https://github.com/exercism/xpython/blob/master/exercises/robot-name/robot_name_test.py

I figured I'd make this PR to keep the scala test suite in line with other exercism.io languages. And once I find where these problem docs are hiding, I'll make it explicit in the README as well :)

ricemery:
Yes. It would be nice if the exercises were consistent between the different language tracks :)
The exercise docs are in https://github.com/exercism/x-common.

Thanks

PR #126, opened 2016-08-20
Reconstruct meetup tests to drive implementation

jmtuley:
This version of the test suite starts each method’s tests with a
concrete example, followed by tests that change each parameter in turn,
and finally groups together the remaining tests as a quick check of the
method.

While it's a matter of style/taste/practice, these are the tests I would write as I test-drove an implementation of `Meetup`. The simplest thing to make the first test in each "group" pass is of course to return a hard-coded object; as one adds requirements, the implementation becomes more and more complete.

The "remaining" tests are retained, but grouped into a single example (since by the time that test is run, the implementation _should_ be complete).

**Note**: I do understand that this change may be seen as a matter of personal taste and get rejected on that basis.


kytrinyx:
Thanks @jmtuley I really like this change.

Issue #131, opened 2016-08-29
pangram: tests not pending


Cohen-Carlisle:
@Trevoke reported:
> Just pulled the pangram exercise for Scala and no tests are pending.

ricemery:
What does this mean? Does this refer to the fact that none of the tests are marked to be ignored?

Trevoke:
Yes. The standard in exercism is to have the first test run and the other tests marked pending so they don't muddy up the test output, letting the user decide when to start running each test. This exercise does not follow this.

ricemery:
This is a bigger problem than just the Pangram exercise. The majority of the Scala exercises do not use the "pending" scheme. It looks like all the exercises from "hello-world" to "gigasecond" use pending. Then all the rest do not use pending. About 46 exercises will need to be updated.

So, addressing this issue will take some time.

Thanks.

Trevoke:
I'm going to open a new issue that reflects the situation better and has a list of unfixed exercises. Please close this one.

PR #132, opened 2016-08-29
Add "pending" to tests in Pangram

Trevoke:
Work on #131


Trevoke:
If this is the right way to fix it, then I can help.

ricemery:
Addresses #131 for Pangram - but the issue should be left open to deal with issue on other exercises.

PR #157, opened 2016-09-23
exercise lense-person

abo64:
I decided to use the Monocle library for the example. It seems to be the only lens library that also supports `Iso` (which is the intended way to implement `setBirthMonth`).

### DIFF
--- a/config.json
+++ b/config.json
@@ -65,7 +65,8 @@
     "parallel-letter-frequency",
     "acronym",
     "zipper",
-    "forth"
+    "forth",
+    "lens-person"
   ],
   "exercises": [
     {
@@ -443,6 +444,12 @@
 		},
     {
       "slug": "forth",
+      "difficulty": 1,
+      "topics": [
+      ]
+		},
+    {
+      "slug": "lens-person",
       "difficulty": 1,
       "topics": [
       ]
--- a/exercises/lens-person/README.md
+++ b/exercises/lens-person/README.md
@@ -1,0 +1,19 @@
+# Lens Person
+
+Use lenses to manipulate (get, set, modify) immutable objects.
+
+Updating fields of nested (immutable) objects is kind of annoying in Scala. One solution
+is to use [lenses](https://wiki.haskell.org/Lens).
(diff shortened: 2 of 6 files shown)


abo64:
Seems the Travis checks insist on Java 7? Why? Can we perhaps upgrade to Java 8?

ricemery:
I think we should be able to update to Java 1.8. I can't think of why we should stay on 1.7 - especially since Java 1.8 has been out for so long. We will need to make sure there are no readmes or any text that indicates that 1.7 is the minimum version.

ricemery:
I am working on updating .travis.yml to only include java 1.8.

abo64:
thx @ricemery

ricemery:
I have updated the .travis.yml. 
But, the reason that travis is failing is because lens-person need to be added to config.json. It needs to be added in two places. lens-person should be added to the problems array. And, should be added to the exercises array. The exercises array is a bit tricky in that a difficulty and topics array should be included. You can just default difficulty to 1 and a blank topic array. Or, if you want the list of topics can be found at https://github.com/exercism/x-common/blob/master/TOPICS.txt . Difficultly is an int 1..10 with 10 being the most difficult.

Can you modify config.json and resubmit the pull request?

Thanks @abo64

abo64:
Thank you for the review and advice @ricemery 
I have added the entry in config.json for lens-person. I just took over the respective entry from Haskell for that. But feel free to change it if you disagree or have a better idea.
I also added a solution for Scalaz in example.scala by splitting things up into traits. Not sure if this is a good idea in terms of readability, though. Again, feel free to change it or tell me if you have some better idea.
And I am unsure about build.sbt. I have included both Monocle and ScalaZ. Perhaps I should better omit them or comment them out? But then example.scala doesn't compile anymore?!

ricemery:
Thanks @abo64 ! Looks good.

On the build.sbt question. I have a bit of a mixed opinion on of Monocle and ScalaZ should be in the build.sbt. Lets leave them there since the example.scala needs the. And, since we are looking for a solution using lenses it makes sense to have Monocle at least in there.

As far as the config.json entry, that is fine for now. The config.json entries need to be updated to include the difficulty, and topics. The entry you put in there is basically a placeholder for now.

Thanks

PR #160, opened 2016-09-27
Improved "bank-account" example

Freshwood:
* Make sure we close redundancies
* Make sure we using the features of case classes
* Make sure we using flatMap as functional approach (The monadic version is prefered by the scala guys)

Greetz

Freshwood


ricemery:
Thanks @Freshwood ! Looks good.

PR #161, opened 2016-10-02
attribute the scala icon and mention its copyright

catb0t:
hooray, this is the last step in the penultimate step
in fixing https://github.com/exercism/exercism.io/issues/3112


kytrinyx:
Yay! So close!

PR #164, opened 2016-10-10
Add tests to Leap exercise

rubek:
The former tests made it able to pass with a very faulty implementation
that seems a bit more intuitive if you don't know about leap years; such as:

* Checking for even years except those years divisible by 4
* Checking for milleniums except those centuries divisible by 400


kytrinyx:
:heart:

PR #166, opened 2016-10-15
Add bracket-push exercise

ricemery:


exercises/bracket-push/example.scala
```diff
@@ -0,0 +1,23 @@
+import scala.util.parsing.combinator.RegexParsers
+
+object Brackets extends RegexParsers {
```
abo64: interesting. :-)
I have to admit that RegexParsers would not have come to my mind as a first idea, but simply keeping track of the various brackets by just counting them. your solution is elegant but may be too advanced for some? (or at least mention to use RegexParsers explicitly in the Readme then?).
or perhaps have exercise sgf-parsing instead for this kind of solution?

ricemery: Thanks for reviewing the solution.

It is probably a bit complex of a solution for the problem at hand. I wrote the example this way because I was thinking that just parsing for parens/brackets was not a real world problem. The real world problem would more likely be to perform some sort of expression or language parsing. The parens are just providing the scope. I figured that an example using parser combinators is more interesting than just showing how to parse using stacks or recursion.
Additionally I was trying not to just use recursion or iteration with a stack to solve the problem. 

I think I would rather nor provide a hint in the readme. I have no intent to force this solution on people. 

We could replace the solution with something more simple. Or, we could provide an alternate solution as another class. 

Here is a simple solution I wrote in Rust. I was trying not to just duplicate the iteration logic in Scala.
```rust

pub struct Brackets {
    s: String
}

impl Brackets {
    pub fn from(s: &str) -> Self {
        Brackets { s: s.to_string()}
    }

    pub fn are_balanced(self) -> bool {
        let mut stack = Vec::new();

        for c in self.s.chars() {
            match c {
                '(' | '[' | '{' => stack.push(c),
                ')' => if stack.is_empty() || stack.pop().unwrap() != '(' {
                            return false;
                        },
                ']' => if stack.is_empty() || stack.pop().unwrap() != '[' {
                            return false;
                        },
                '}' => if stack.is_empty() || stack.pop().unwrap() != '{' {
                            return false;
                        },
                _ => continue,
            }
        }

        return stack.is_empty();
    }
}```

abo64: As I said, I like your solution very much (it even motivates me look into Parsers and sgf-parsing). :-)
I was (and still am) not quite sure about the exact role of these examples, though. People are not supposed to ever look at it, so is it more for making sure that the TestSuite is correct?
In this case you definitely can have some fun and not just repeat some standard solution you already know.  
In summary: I am totally fine with your solution. :-)

ricemery: I just went searching around the site documentation and found the following -

> Reference Solution

> The reference solution is named something with example or Example in the path.

> The solution does not need to be particularly great code, it is only used to verify that the  exercise is coherent.

So, it sounds like having an overly complex example is fine:)

Thanks again for your feedback!

> If you change the test suite, then make sure the reference solution is fixed to pass the  updated tests.

PR #212, opened 2016-10-28
alphametics: added new tests, improved example

abo64:
Modified and added tests according to new `x-common/exercises/alphametics/canonical-data.json`.
Actually they removed the tests for multiplication and exponents, but I decided to keep them as they make the exercise more interesting.

### DIFF
--- a/exercises/alphametics/example.scala
+++ b/exercises/alphametics/example.scala
@@ -16,23 +16,14 @@ object Alphametics {
     ExpressionParser.parse(str)
 
   private def solveExpression(uniqueLetters: String)(boolExpr: Expression[Boolean]): Option[Solution] = {
+    def isSolution(solution: Solution): Boolean =
+      boolExpr.eval(solution) getOrElse false
+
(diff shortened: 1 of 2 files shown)


ricemery:
Thanks @abo64 .

As an FYI - When I have been adding new exercises I have been creating Scala code to reading in the canonical json and spit out the unit tests. The code isn't all that great. I haven't spent any time making it nice. On limitation is that each new exercise needs a new generator - as the json and interface under test is different for each.
Another limitation is that I am not spitting out formatted scala. I use  IntelliJ to reformat the tests after they are spit out.

My intent was that it should be easy to regenerate the unit tests if the json changes. And, It saves me from having to write the unit tests by hand.

The generators are in the testgen/src/main/scala directory.

Anyways, don't feel like you have to do the same. Just wanted to make you aware of my process.

-Ric

kytrinyx:
@ricemery - would you add a note about this to the README?

ricemery:
Yep. I will add a note.

Issue #220, opened 2016-11-29
Add hints to some exercises


abo64:
Most exercises have their own favorable solutions and characteristic pitfalls.
To give people some guidance and avoid writing repetitive or similar code comments again and again
we should add a `HINTS.md` file to any exercise where it makes sense.

Especially the first couple of exercises should have such a file to assist beginners.
Another case is exercises dealing with special features of the language like `Option` or `Either`.

ErikSchierboom:
Great idea. I would also suggest to include hints for the text-parsing exercises to Scala's parser combinator feature. 

For the `Option` and `Either` hints, we could link to the Neophyte's Guide to Scala, which has an excellent explanation of those topics.

abo64:
@ErikSchierboom I am working on `Option` and `Either` hints, and I also love "Neophyte's Guide to Scala"! ;-)
Hope to have the PR ready soon.

@ErikSchierboom  @ricemery Btw, first I intended to add the whole list of exercises to this issue so that we can checkmark those containing hints, but decided against it because I thought we will have hints only for a couple but not all of them. Or do you disagree and we should add this list?

ErikSchierboom:
I don't think we should do this for all exercises, but only for those where it makes sense. Also, in general, the smaller the PR the better, so my preference would be to create one PR per exercise.

Creating a list for all exercises that need hints might be wise, as it allows us to keep track of progress.

abo64:
thx @ErikSchierboom 
I will add a such a list, and you guys feel free to add/modify it.
And let's not forget to keep it up to date.

Issue #214, opened 2016-11-01
BankAccount: not scala-ish enough?


Trevoke:
This is a very Java-ish test suite and the implementation follows. There are probably better ways to implement this in Scala... Should the test suite be updated?

ricemery:
What specifically is not Scala-ish? The exercise is testing an object with mutable state. How would you suggest this be done in a different manner? 

I could see the exercise being written/solved using Actors. But, it seems like developers should know how to mutate state in a threaded environment without resorting to Actors.

Thanks!

Trevoke:
Well, the option of not using mutable state, for instance -- or being able to return a new object every time.

Frankly I barely even registered the threads, since that need is contained to a very specific part of the system. I was referring to the behavior of open account / closed account, which, given the test suite, we can't really handle elegantly with type systems (or at least, *I* can't, because I've only been working in Scala for a couple of months). I went through [this blog entry](https://ktoso.github.io/scala-types-of-types/) on types and couldn't find something that would be helpful here. Shadow types were *almost* helpful but didn't work.

ricemery:
Thanks @Trevoke 

I agree that returning new immutable instances for each account operation is cleaner. But, I think the original intent of this exercise was to specifically force the implementer to deal with mutable state. At some point within a banking app mutable state must be dealt with..

I went and looked at the Haskell and Clojure examples for the same exercise. Both are also using mutable state within the sample solutions. So, I think the intent is right.

Looking at the scala example and tests, I do think that the fact that creating a new instance results in "opening" an account is odd. An openAccount function would probably make more sense.

I will be interested to see if anyone else chimes in on this issue report.

ricemery:
@abo64 , @ErikSchierboom 

Gentlemen,

What do you think of this issue?

Thanks

abo64:
I also think that currently this exercise is about mutable state.
Although it might be an interesting idea to add another exercise for such an FP-style immutable bank account. Would be of great educational value.

But for this exercise we could indeed have a factory method `BankAccount.open` or so.

ErikSchierboom:
I also feel that this exercise is about mutable state, and as such introduces people to the hybrid design of Scala. Doing a pure FP-style exercise would be nice, but I think the exercise is good the way it is at the moment. A factory method `BankAccount.openAccount` wouldn't hurt of course.

ricemery:
[...]

Issue #137, opened 2016-09-15
Exercises do not provide a named file for the code to be delivered


Trevoke:
In other tracks, there is a file that people can use to write their code (whether `bob.el` or `bob.rb` or `bob.ex`). In the Scala track, currently, that file does not exist, so we have to guess what the file name is supposed to be.

This issue tracks the exercises that still need to be updated. No checkmark means the exercise still needs to be updated.
- [ ] Hello World
- [x] Bob
- [ ] Hamming
- [ ] Word Count
- [x] Pangram
- [ ] Anagram
- [ ] Nucleotide Count
- [ ] Phone Number
- [x] Grade School
- [ ] Robot Name
- [ ] Leap
- [ ] Etl
- [ ] Meetup
- [x] Space Age
- [ ] Grains
- [x] Gigasecond
- [ ] Scrabble Score
- [x] Difference Of Squares
- [x] Accumulate
- [ ] Sublist
- [ ] Rna Transcription
- [ ] Triangle
- [ ] Binary
- [ ] Roman Numerals
- [ ] Prime Factors
- [x] Raindrops
- [x] Allergies
- [x] Atbash Cipher
- [x] Bank Account
- [x] Crypto Square
- [ ] Kindergarten Garden
- [ ] Robot Simulator
- [x] Queen Attack
- [ ] Binary Search Tree
- [ ] Hexadecimal
- [ ] Largest Series Product
- [ ] Luhn
- [ ] Clock
- [ ] Matrix
- [ ] House
- [ ] Minesweeper
- [ ] Ocr Numbers
- [ ] Octal
- [ ] Trinary
- [ ] Wordy
- [ ] Food Chain
- [ ] Linked List
- [ ] Custom Set
- [ ] Nth Prime
- [ ] Palindrome Products
- [ ] Pascals Triangle
- [ ] Pig Latin
- [ ] Pythagorean Triplet
- [ ] Saddle Points
- [ ] Say
- [ ] Secret Handshake
- [ ] Series
- [ ] Sieve
- [ ] Simple Cipher
- [ ] Connect
- [x] Parallel Letter Frequency
- [x] Acronym
- [x] Forth

ricemery:
It doesn't necessarily matter what the class file is named. If the user wants the class file named to match the class used by the tests, then this is easy enough to deduce from the tests. 
Unless we include stubbed out classes and objects, the user is going to have to figure out the class name from the tests anyways.

[...]

PR #221, opened 2016-11-30
Rework interface definition to cleanly show openAccount. Add HINTS.md…

ricemery:
1) Rework interface definition to cleanly show openAccount. 
2) Add HINTS.md with note about mutable state. 
3) Add Bank.scala with BankAccount trait that should be implemented, and Bank object that should be filled out by the implementer with BankAccount construction.
Refs #214, #220, #137


exercises/bank-account/HINTS.md
```diff
@@ -0,0 +1,4 @@
+## Hints
+
+This exercise is testing mutable state that can be accessed saftely from multiple threads. Scala provides a variety of ways to protect 
+mutable state. For developers familiar with Java concurrency, Scala can utilize the Java concurrency support such as the Java synchronized block.
```
abo64: Although this might already be some sort of spoiler you could add something like:
```
In Scala there are two ways to achieve mutable state: Use a "var" or a mutable object.
Two common mistakes here are:
- Do not use a "var" that is also a mutable object. One is enough, but not both together.
- Don't expose the "var" or mutable object to the outside world. So make them "private" and change the mutable object into immutable before you return it as a value.
```
One could also imagine a standard format with a separate section named "Common Pitfalls" or so.

exercises/bank-account/src/main/scala/Bank.scala
```diff
@@ -0,0 +1,13 @@
+trait BankAccount {
+
+  def closeAccount(): Unit
+
+  def getBalance: Option[Int]
+
+  def incrementBalance(increment: Int): Option[Int]
+}
+
+object Bank {
```
abo64: Why don't you want to use the companion object anymore?
Same for the file name. AFAIK the file name is always the same as the class/object to be implemented?

exercises/bank-account/HINTS.md
ErikSchierboom: The suggested addendum by @abo64 seems like a good one, even though it is a bit of a spoiler.

exercises/bank-account/src/main/scala/Bank.scala
ErikSchierboom: Well, there is a `Bank` object being implemented of course, so that sort of makes sense. I usually named the files after the slug name, so for the `BankAccount` exercise, I would use `BankAccount.scala`.

ricemery: I changed the name for the Object from BankAccount to Bank. I didn't like having a BankAccount.openAccount function. It seems like a Bank should be opening accounts not a BankAccount.  I thought about having a Bank.scala and BankAccount.scala. But, then I didn't like having 2 .scala files for the exercise..

ricemery:
I renamed Bank.scala to BankAccount.scala . At least now the source file is named to match the test cases file.
I also added the "Common pitfalls" text to the HINTS.

Thanks!

ricemery:
Thanks!

PR #222, opened 2016-11-30
Add hints to hello-world exercise

abo64:
#220 add `HINTS.md` to `hello-world` exercise.
Question is whether we should agree on a common format for those hints?

### DIFF
--- a/exercises/hello-world/HINTS.md
+++ b/exercises/hello-world/HINTS.md
@@ -1,0 +1,8 @@
+## Hints
+For this exercise two Scala features come in handy:
+- [Default Parameter Values](http://docs.scala-lang.org/tutorials/tour/default-parameter-values.html)
+- [String Interpolation](http://docs.scala-lang.org/overviews/core/string-interpolation.html)
+
+#### Common pitfalls that you should avoid
+- `null` is usually not considered a valid value in Scala, and there are no `null` checks needed (if you don't have to interface with Java code, say). Instead there is the [Option](http://danielwestheide.com/blog/2012/12/19/the-neophytes-guide-to-scala-part-5-the-option-type.html) type if you want to express the possible absence of a value. But for this exercise just assume a normal non-`null` String parameter.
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
I love this. The only thing I'm not sure of, is the link for the no-null hint. It now links to Wikipedia, but I think a Scala-specific link would be better. I've managed to find [this explanation](https://tpolecat.github.io/2014/05/09/return.html), which does are fairly good job of explaning *why* you shouldn't use `return`. It might be a bit difficult for newbies though.

abo64:
@ErikSchierboom I was also not too happy about this link, but couldn't find anything nice (incl. your link which I also considered too theoretical for beginners).
Now I just searched again and stumbled upon [this](http://stackoverflow.com/questions/24856106/return-in-a-scala-function-literal). What do you think?

ErikSchierboom:
Much better. I think it is at least good enough until we find something better.

abo64:
changed the link.

abo64:
@ErikSchierboom Seems it cannot handle these ordered lists.
Changed it to an unordered list (using hyphens).

ErikSchierboom:
Looks good. 

@ricemery would you like to merge the PR's yourselves, or should we (the other maintainers) also be allowed to do that? Perhaps we should make it a rul that we cannot merge our own PR's, to prevent overlooking mistakes.

abo64:
I found [this](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#maintaining-a-track) quite plausible. But:
*Avoid merging your own pull requests (but it's fine if it's really simple)* - the question is always what is "simple".

ErikSchierboom:
Indeed. But now that we have several maintainers, I think we could try to not merge our own PR's. But I don't know what the others thing about this.

abo64:
Most often the rules without any exceptions are the best.
So I also think we should try to never merge our own PRs.

ricemery:
Agreed. By default we should not merge our own PRs.

ErikSchierboom:
@ricemery So it would be okay if I merged this PR?

ricemery:
Absolutely!

Thanks

ErikSchierboom:
Thanks @abo64!

PR #224, opened 2016-11-30
Add hints to hamming exercise

abo64:
#220 add `HINTS.md` to `hamming` exercise.
I tried to be as close as possible to the hints for `Either`. Now is this good or bad?
What about the other exercises that use `Option`? Perhaps hints for the first one is enough?


ErikSchierboom:
Once again, great hints. I think only having the hints for the first one is good enough.

PR #226, opened 2016-12-02
Add hints to phone-number exercise

abo64:
#220 
Now that we have decided to only have hints for the first exercise using `Option` one could also opt that this one is not worth the effort?


exercises/phone-number/HINTS.md
```diff
@@ -0,0 +1,8 @@
+## Hints
+For simplicity and readability: Consider using the Scala collection functions instead of Java's `String` methods. Remember that in Scala is implicitly also a `Seq[Char]`, so you can call them as easily as the `String` methods.
```
ErikSchierboom: `Remember that in Scala` should be `Remember that in Scala a string`

ErikSchierboom:
I think these hints are fine. They are helpful and don't give away too much.

abo64:
thx @ErikSchierboom

PR #227, opened 2016-12-05
Add hints to leap exercise

abo64:
#220

### DIFF
--- a/exercises/leap/HINTS.md
+++ b/exercises/leap/HINTS.md
@@ -1,0 +1,4 @@
+## Hints
+Try to avoid code repetition, use private helper functions if you can.
+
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
I wonder if people will understand the "single Boolean expression" bit. Perhaps we could link to an example?

abo64:
Best link I could find is [this one](http://cs.wellesley.edu/~cs111/spring00/lectures/boolean-simplification.html). Unfortunately it is in Java, but might still be OK?

ricemery:
I think it would be fine to include the link.

ErikSchierboom:
Agreed, I think it's a good hint.

abo64:
Added the link.

ErikSchierboom:
I think this looks fine. Merged.

PR #229, opened 2016-12-05
Add hints to gigasecond exercise

abo64:
#220 
I was wondering if we should not even check in the test suite that the original `GregorianCalendar` has not been modified?


ErikSchierboom:
There is nothing in the [canonical data](https://github.com/exercism/x-common/blob/master/exercises/gigasecond/canonical-data.json) that says anything about immutability, so fine by me.

Issue #230, opened 2016-12-07
Use Sets in sum-of-multiples?


abo64:
I just had [this little discussion](http://exercism.io/submissions/f2c92a94a3d346c9bd6278fe456a0cc6).
@enricospataro may have a point here?! Shall we change the test suite?

ErikSchierboom:
Hmmm, he has a point. The readme states: "Write a program that can find the sum of the multiples of a given **set** of numbers." Perhaps this might be a good exercise to get users to practice with sets? We could then add a hints file to help ease them into using sets.

abo64:
I had a quick look, [in the Haskell track](http://exercism.io/exercises/haskell/sum-of-multiples/test-suite) they also use lists. So this might be something bigger.  
@ErikSchierboom But yes, we could change our test suite and add the hints.

ErikSchierboom:
Well, the Haskell track uses lists for almost everything :)

abo64:
@ErikSchierboom True, and that may be because only lists are directly supported in `Prelude`. :)

enricospataro:
It also doesn't make sense to find multiples of duplicate numbers, so I think that Set is the right choice

ricemery:
I think we should change the test suite to use a Set.. It is right there in
the Readme - "Write a program that can find the sum of the multiples of a
given set of numbers."

PR #232, opened 2016-12-10
Change function signature to use Set instead of List. Add template ob…

ricemery:
Change function signature to use Set instead of List. Add template object definition. Add Sets to congig.json topics for exercise. Refs #230


ErikSchierboom:
Great change! Thanks. 🎉

PR #233, opened 2016-12-12
Add hints to parallel-letter-frequency exercise

abo64:
#220 
**Edit**: Perhaps it gives away too much? Should we only have the part about `Future`?


ErikSchierboom:
I think it does give away quite much, but if people usually struggle with this exercise, I don't think that's necessarily bad.

PR #236, opened 2016-12-12
Add test that a large number of new robots receive unique names.

stevejb71:
I also made a few formatting changes (IntelliJ was warning about unnecessary brackets).

### DIFF
--- a/exercises/robot-name/example.scala
+++ b/exercises/robot-name/example.scala
@@ -1,21 +1,18 @@
 import scala.util.Random
 
-class Robot {
-  private var _name: String = makeName
-
-  def name: String = _name
-
-  def reset(): Unit = _name = makeName
-
-  private def makeName = prefix + suffix
-
-  private def prefix = Random.shuffle(Robot.alphabet).take(2).mkString
-
-  private def suffix = Random.shuffle(Robot.suffixes).head.toString
+object UniqueNames {
+  val names =
+    Random.shuffle(1 to 26 * 26 * 1000).iterator.map(n => {
+      val lettersPart = n / 1000
+      val firstLetter = 'A' + (lettersPart / 26)
+      val secondLetter = 'A' + (lettersPart % 26)
+      f"$firstLetter%c$secondLetter%c${n % 1000}%03d"
+    })
 }
 
-object Robot {
-  private val alphabet: Seq[Char] = 'A'.to('Z').toList
+class Robot {
+  private var storedName = UniqueNames.names.next()
 
-  private val suffixes: Seq[Int] = 100.to(999).toList
+  def name: String = storedName
+  def reset(): Unit = storedName = UniqueNames.names.next()
 }
--- a/exercises/robot-name/src/test/scala/RobotNameTest.scala
+++ b/exercises/robot-name/src/test/scala/RobotNameTest.scala
@@ -1,10 +1,12 @@
 import org.scalatest._
 
+import collection.mutable
+
 class RobotNameSpecs extends FunSpec with Matchers {
   val nameRegex = """[A-Z]{2}\d{3}"""
 
   it ("has a name") {
-    new Robot().name should fullyMatch regex (nameRegex)
+    new Robot().name should fullyMatch regex nameRegex
   }
 
   it ("does not change its name") {
@@ -16,7 +18,7 @@ class RobotNameSpecs extends FunSpec with Matchers {
 
   it ("does not have the same name as other robots") {
     pending
-    new Robot().name should not be (new Robot().name)
+    new Robot().name should not be new Robot().name
   }
 
   it ("can have its name reset") {
@@ -25,7 +27,22 @@ class RobotNameSpecs extends FunSpec with Matchers {
     val name = robot.name
     robot.reset()
     val name2 = robot.name
-    name should not equal (name2)
-    name2 should fullyMatch regex (nameRegex)
+    name should not equal name2
+    name2 should fullyMatch regex nameRegex
+  }
+
+  // Making this test pass is an optional extra exercise, should you want more of a challenge.
+  // It's ignored by default, to make it run, simply change "ignore" below to "it".
+  // There are 26^2 * 1,000 = 676,000 possible robot names - you have to ensure that none are repeated.
+  // The Robot code needs to be efficient enough to allow all 676,000 unique names to be generated.
+  ignore("a large number of new instances have unique names") {
+    val alreadySet = mutable.HashSet.empty[String]
+    for(_ <- 0 until 676000 - 6) { // as 6 robot names are generated in the tests above!!
+      val name = new Robot().name
+      if (alreadySet contains name) {
+        fail(s"$name is repeated")
+      }
+      alreadySet += name
+    }
   }
 }


abo64:
@ErikSchierboom @ricemery What do you think?
It seems to run fast enough, but it is (and cannot be?) included in `x-common`?!

ErikSchierboom:
Well, I think the goal of the added test is great, but I'm unsure about the number of iterations it checks. I think just testing for 100 or 1000 iterations should be more than enough for the purposes of the exercise.

The x-common entry states:

> In some exercism language tracks there are tests to ensure that the same name is never used twice.

That also leads me to think that this is not the most essential property of the exercise. All in all, I think the addition is good, but the number of iterations should be reduced. Who knows what slow algorithm people can come up with. I would not like the test to stop those solutions from being submitted.

stevejb71:
1,000 is a little low (the tests falsely pass sometimes) - 10,000 should be enough though.

ricemery:
I went searching through some other language test implementations. Haskell,
Clojure, Java, Ruby and Rust do not implement this test case. PHP has the
test case using 10,000.

I don't have a strong opinion. Though I am not sure I understand why 10,000
would be the magic number to test. If there are (26 ^ 2) * 1000 possible
names, that would be 676000 possible names. Which means that only 1% of the
possible names would be tested if 10000 iterations was used. In no way
would I suggest testing all the names. Just pointing out that testing 1% is
probably not any better than testing .5% or .1%.
Is the 10,000 number somehow statistically significant in testing the
possible names?

stevejb71:
You're right, 10,000 is way too low. 

Checking all 676,000 would be the only way to guarantee no false positives. How about putting this test in a separate spec, pending, with a comment on it being an optional extra challenge?

abo64:
We had some problematic tests in a Haskell exercise. The solution we agreed on [there](https://github.com/exercism/xhaskell/blob/master/exercises/alphametics/test/Tests.hs#L63) was to comment them out together with an explanation.
Maybe that could be a feasible compromise here, too?

ErikSchierboom:
Summarizing, I think we have two issues:

1. What is the correct number of iterations to use to test for uniqueness?
2. Should we include the test by default or not.

If the answer for option 2 is: they are not included by default, but are in comments, I'm fine with leaving the number in 1 a bit high.

stevejb71:
I am in favour of following what @abo64 said was done in the Haskell stream, and leave the test in there but disabled by default, as an optional exercise. Rather than comment out (and leave the code possibly to rot) it can be just left pending, with a comment.

In that case, the test might as well check all 676,000 robot names are unique. The exercise implementor will have to find an efficient solution, but that can be part of the challenge.

abo64:
@stevejb71 But then I would rather set it to `ignore` instead of `pending` plus an explaining comment.
All other tests but the first are initially set to `pending`, and `ignore` would make the distinction more explicit.

ErikSchierboom:
I agree with the `ignore` suggestion. Better to make it distinct from the existing `pending` tests.

ricemery:
I agree with the ignore suggestion as well.

stevejb71:
I added a new ignored test - this checks that there are no repeats when all possible names are generated. Because global state is used to keep the uniqueness checking, and previous tests use up some names, slightly fewer than 676,000 names are generated. It's not very pretty.

The existing solution fails the new test, so that is also replaced.

exercises/robot-name/example.scala
```diff
@@ -1,21 +1,18 @@
...
-
-  private def makeName = prefix + suffix
-
-  private def prefix = Random.shuffle(Robot.alphabet).take(2).mkString
-
-  private def suffix = Random.shuffle(Robot.suffixes).head.toString
+object UniqueNames {
+  val names =
+    Random.shuffle(1 to 26 * 26 * 1000).iterator.map(n => {
+      val lettersPart = n / 1000
```
abo64: Does this really generate all possible names, or aren't some `n`s mapped to the same name?
But well, if your new test succeeds then that's the answer ... :)

exercises/robot-name/src/test/scala/RobotNameTest.scala
```diff
@@ -1,31 +1,45 @@
...
 
   it ("does not change its name") {
-    pending
     val robot = new Robot
     val name = robot.name
     robot.name should be (name)
   }
 
   it ("does not have the same name as other robots") {
-    pending
```
abo64: Please keep the `pending`s.
This is how all test suites are designed: All but the first test are set to `pending`.

exercises/robot-name/src/test/scala/RobotNameTest.scala
```diff
@@ -1,31 +1,45 @@
...
-    new Robot().name should not be (new Robot().name)
+    new Robot().name should not be new Robot().name
```
abo64: I guess the parentheses are there because most or all tests have them after `be`.
But if it compiles then personally I don't mind removing them.

exercises/robot-name/src/test/scala/RobotNameTest.scala
```diff
@@ -1,31 +1,45 @@
...
+
+  // Making this test pass is an optional extra exercise, should you want more of a challenge.
+  // It's ignored by default, to make it run, simply change "ignore" below to "it".
+  // There are 26^2 * 1,000 = 676,000 possible robot names - you have to ensure that none are repeated.
+  // The Robot code needs to be efficient enough to allow all 676,000 unique names to be generated.
+  ignore("a large number of new instances have unique names") {
+    val alreadySet = mutable.HashSet.empty[String]
+    for(_ <- 0 until 676000 - 6) { // as 6 robot names are generated in the tests above!!
+      val name = new Robot().name
+      if (alreadySet contains name) {
```
abo64: or
```scala
if (alreadySet(name)) {
```

exercises/robot-name/src/test/scala/RobotNameTest.scala
stevejb71: Oops, I was running them to test. I'll add them back in when I'm online next.

exercises/robot-name/src/test/scala/RobotNameTest.scala
stevejb71: You need them with should be, but not with should not be :).

abo64:
@ricemery @ErikSchierboom 
Anything to add from your side?
Or can we merge?

ErikSchierboom:
Nothing to add. LGTM.

ricemery:
I think in general I would never write a unit test to test all the possible
outputs of a function with such a large output set. I suspect that there is
a mathematically/statistically valid way to test the function without
resorting to looping over all the outputs. Understanding how to test
collisions generated by a Hash function would probably identify how to test
the the robot name function. But, this is in no way my area of expertise.

So, I say we go ahead and merge the change.

abo64:
@ricemery I had similar thoughts, too. This only works because the number of possible outputs is still small enough. I always tell people in my reviews to generalize their code for larger name sizes, and that might then not be so easy anymore.
On the other hand, the new test case is just an additional advanced challenge as stated in the accompanying comment.

Fine then. I think I made a mistake last time by just clicking on "Merge pull request".
What is the right way to merge in order to get the right commit message? In this case probably "Squash and merge"?

ricemery:
@abo64 I just did a merge without the squash. I think it looks fine.

PR #241, opened 2016-12-13
Renamed test file names to PascalCase

### LINKED ISSUE #237: Change test file names to match Scala convention.
For example, in robot-name, the filename for the spec is robot_name_test.scala.

Underscores are unusual in Scala filenames. I'd expect something like RobotNameSpec.scala.

ErikSchierboom:
This PR fixes #237.


ricemery:
Thanks @ErikSchierboom !

Issue #242, opened 2016-12-14
Use an object instead of a class for some exercises


abo64:
Some exercises require a `class` even though they don't have any internal state.
So they should be changed into an `object`.

- [x] allergies
- [x] atbash-cipher
- [x] crypto-square
- [x] queen-attack
- [x] raindrops

ricemery:
@abo64 , I can work through these or a subset of them if you haven't already started.

I do think that we should also address #137 for these exercises as long as we are modifying the exercises.

abo64:
@ricemery No, I haven't started. Please go ahead!

PR #243, opened 2016-12-15
Allergies - use object instead of an instance - refs #242. Add Allerg…

ricemery:
Allergies - use object instead of an instance - refs #242. 

Add Allergies.scala to src dir - refs #137. I don't really like the template I put in the file for the Allergens enumeration. But, I think it at least points out that the enumeration must be created.

### DIFF
--- a/config.json
+++ b/config.json
@@ -173,6 +173,9 @@
       "slug": "allergies",
       "difficulty": 1,
       "topics": [
+         "Lists",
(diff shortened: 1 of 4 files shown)


ErikSchierboom:
Although the template for enumeration is not the prettiest, I think it *will* help people get started. Thanks a lot! 🎉

PR #244, opened 2016-12-15
Atbash - use object instead of an instance - refs #242. Add Atbash.sc…

ricemery:
Atbash - use object instead of an instance - refs #242.
Add Atbash.scala - refs #137. 


ErikSchierboom:
Thanks! 🎉

PR #245, opened 2016-12-15
CryptoSquare - use object instead of an instance - refs #242. Add Cry…

ricemery:
CryptoSquare - use object instead of an instance - refs #242. 

Add CryptoSquare.scala - refs #137. 


ErikSchierboom:
Merged! 🎉

PR #246, opened 2016-12-15
Queens - use object instead of an instance - refs #242. Add Queens.sc…

ricemery:
Queens - use object instead of an instance - refs #242. 

Add Queens.scala - refs #137. 


abo64:
thanks @ricemery !

PR #247, opened 2016-12-15
Raindrops

ricemery:
Raindrops - use object instead of an instance - refs #242. 

Add Raindrops.scala - refs #137. 


ErikSchierboom:
Merged! 🎉

Issue #235, opened 2016-12-12
Upgrade to Scala 2.12 ?


abo64:
Shall we upgrade to Scala 2.12, perhaps in Q1 2017 ?
This includes ScalaTest and other libraries like Monocle, of course.
A problem could be `scala-parser-combinators`?

Would be great if we find a way to simplify this for the future, ideally by just having to change only a single file?

ErikSchierboom:
I think we should. The main argument for me is that when people search for documentation, they will be led to the latest version's documentation. Ideally, our track should match that.

Weren't the parser combinators extracted to their own library? If so, we can just add the reference to the SBT config file, right?

ricemery:
The parser combinator dependency has already been broken out - at least in
bracket-push. I don't recall what other exercises use the the lib. But,
this was an issue I had to deal with when we moved to 2.11.

ricemery:
What order should we change the exercises to 2.12? I am thinking we should start at the last exercise in the series and move backwards to Hello World.

abo64:
I just saw that [parser-combinators](https://github.com/scala/scala-parser-combinators) also has a 2.12 version, so it should be smooth sailing.

@ricemery Yes, moving backwards sounds like a good strategy.

Is there a way to use variables that are read from a single source in the root directory or so? Or do we have do go through all exercises again for the next major upgrade?

ErikSchierboom:
> Is there a way to use variables that are read from a single source in the root directory or so? Or do we have do go through all exercises again for the next major upgrade?

Unfortunately, I don't think this is possible at the moment. There is a long-standing [open issue](https://github.com/exercism/cli/issues/88) on the CLI that would add support for this feature. It would be great if someone would implement that, but I don't know any Go unfortunately. The F# and C# tracks could also really benefit for it.

abo64:
@ricemery Just noticed that in Scala 2.12 `Either` is finally right-biased which should simplify things considerably.
So we might give the respective exercises a higher priority:
- nucleotide-count
- bowling
- forth

ricemery:
I was just thinking about starting to porting exercises over to 2.12. I'll
start with these three.. I'll try to get started tonight if possible.

ricemery:
TODO

- [ ] Hello World
- [ ] Bob
- [ ] Hamming
- [ ] Word Count
- [ ] Pangram
- [ ] Anagram
- [ ] Nucleotide Count
- [ ] Phone Number
- [ ] Grade School
[...]

PR #253, opened 2016-12-26
Update to scala 2.12.1 - refs #235

ricemery:
Update to scala 2.12.1 - refs #235


abo64:
thank you @ricemery !

Issue #259, opened 2016-12-30
gigasecond: Switch to java.time library


abo64:
The [Java track](http://exercism.io/exercises/java/gigasecond/test-suite) has switched from the clunky `java.util.GregorianCalendar` to `java.time.LocalDateTime`.
I think we should do the same?

Another question might be whether we should not change `Gigasecond` to an `object` as it is not really stateful?

PR #262, opened 2017-01-04
gigasecond: Switch to java.time library

abo64:
#259 Use `java.time` library
#235 Upgrade to Scala 2.12
#125 Add topics


ricemery:
Thanks @abo64 !

PR #263, opened 2017-01-05
new exercise run-length-encoding

abo64:
As always I am not sure where to put this exercise in `config.json`.
Assuming we keep this exercises near to the top I provided a stub with the function signatures instead of an empty `.keep` file.


ricemery:
Cool.  Thanks @abo64

PR #265, opened 2017-01-05
roman-numerals: Change class to object (issue #242)

abo64:
#242 Change to object
#235 Upgrade to Scala 2.12
#125 Add topics to `config.json`


ricemery:
Thanks @abo64

PR #282, opened 2017-01-16
Trinary - Update to Scala 2.12.1 - refs #235.

ricemery:
Update to Scala 2.12.1 - refs #235.

Note that Trinary is actually deprecated.. So, there probably wasn't much reason to update the build.sbt. But it was easy to do...


abo64:
thanks @ricemery

PR #305, opened 2017-01-24
Scrabble Score - Update to Scala 2.12.1 - refs #235. Add topic…

ricemery:
Update to Scala 2.12.1 - refs #235. 
Add topics to config.json - refs #125. 
Change solution to use object instead of class.


abo64:
thanks @ricemery

Issue #319, opened 2017-01-26
Move deprecated exercises out of exercises dir


ricemery:
We should move the deprecated exercises out of the exercises dir and into some other sub dir of xscala. Maybe just name the dir deprecated. 

This way we can keep the deprecated exercises around, without having them polluting the live exercises dir.

ricemery:
Can't merge. This would break exercism.

PR #328, opened 2017-01-28
Move deprecated exercises out of the exercises dir.

ricemery:
Move deprecated exercises out of the exercises dir. Refs #319


ErikSchierboom:
@ricemery @kytrinyx Is this safe to do? What about people that have already fetched the deprecated exercises and try to update. Will that fail with a nice error message?

kytrinyx:
This is not safe to do. Moving them out of exercises will make them unavailable as soon as we delete the code that handles the duplication.

ricemery:
Bummer. :(

PR #332, opened 2017-02-02
Add some detail to INSTALLATION.md. And, update required Java version…

ricemery:
Add some detail to INSTALLATION.md. 
And, update required Java version.
Refs #248

This is modeled off Java INSTALLATION.md. Though much simplified.

I looked at the Java version and the C# version. In both cases they go into great detail on how to install the needed tools. This seems really heavy weight to me. I would rather redirect users to other resources for hints on how to install Java, sbt, and IDEs. Documenting this stuff ourselves seems like we would be duplicating effort best done by Java, sbt and IDE teams..


abo64:
thanks @ricemery

PR #333, opened 2017-02-03
Add binary-search exercise

ricemery:
Add binary-search exercise

### DIFF
--- a/exercises/binary-search/example.scala
+++ b/exercises/binary-search/example.scala
@@ -1,0 +1,23 @@
+import scala.annotation.tailrec
+
+object BinarySearch {
+  def search[T](seq: Seq[T], value: T)(implicit ord: T => Ordered[T]): Option[Int]
+    = searchInternal(seq, value, 0, seq.size - 1)
+
+  @tailrec
+  def searchInternal[T](seq: Seq[T], value: T,
+                        start: Int, end: Int)(implicit ord: T => Ordered[T]): Option[Int] = {
+    if (end < start || start < 0)
+      None
+    else {
+      val middle = Math.floor(0.5 * (start + end)).asInstanceOf[Int]
+      val elem = seq(middle)
+      if (elem == value)
+        Some(middle)
+      else if (value < elem)
+        searchInternal(seq, value, start, middle - 1)
+      else
+        searchInternal(seq, value, middle + 1, end)
+    }
+  }
+}
--- a/exercises/binary-search/src/test/scala/BinarySearchTest.scala
+++ b/exercises/binary-search/src/test/scala/BinarySearchTest.scala
@@ -1,0 +1,74 @@
+import org.scalatest.{FunSuite, Matchers}
+
+class BinarySearchTest extends FunSuite with Matchers {
+  test("finds a value in an array with one element") {
+    pending
+    val elements = Array[Int](6)
+    val result = BinarySearch.search(elements, 6)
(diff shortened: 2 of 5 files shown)


exercises/binary-search/src/test/scala/BinarySearchTest.scala
```diff
@@ -0,0 +1,74 @@
+import org.scalatest.{FunSuite, Matchers}
+
+class BinarySearchTest extends FunSuite with Matchers {
+  test("finds a value in an array with one element") {
+    pending
+    val elements = Array[Int](6)
```
abo64: Remove the `[Int]`?

exercises/binary-search/example.scala
```diff
@@ -0,0 +1,23 @@
+import scala.annotation.tailrec
+
+object BinarySearch {
+  def search[T](seq: Seq[T], value: T)(implicit ord: T => Ordered[T]): Option[Int]
```
abo64: I also prefer `Seq` over `Array`.
But does `x-common/binary-search` not even insist on `Array`?
(cf. the comment at the top in `canonical-data.json`).

exercises/binary-search/src/test/scala/BinarySearchTest.scala
```diff
@@ -0,0 +1,74 @@
...
+    val result = BinarySearch.search(elements, 6)
+    assert(result.contains(0))
```
abo64: Why not be more explicit about the return type as in the other exercises?
```scala
result should be (Some(0))
```
(same for the other test cases).

exercises/binary-search/src/test/scala/BinarySearchTest.scala
```diff
@@ -0,0 +1,74 @@
+import org.scalatest.{FunSuite, Matchers}
+
+class BinarySearchTest extends FunSuite with Matchers {
+  test("finds a value in an array with one element") {
+    pending
```
ricemery: Oops. I though I fixed this.

exercises/binary-search/src/test/scala/BinarySearchTest.scala
ricemery: I have the [Int] there because the very last test case did not compile without the hint. I wasn't;t real jazzed about having the [Int] there. But, I didn't want to hand edit the test case to only put it where needed. I suppose I could add more logic in the generator to only put it where needed.

exercises/binary-search/src/test/scala/BinarySearchTest.scala
ricemery: IntelliJ was complaining about Some(0). I forget the actual complaint.
I think I like Some(0) and None better. So, since you agree I'll ignore IntelliJ and go back to Some/None.

exercises/binary-search/example.scala
ricemery: I originally wrote the tests to use Seq. But, all the test case comments from the canonical-data.json use the term Array. It felt wrong to have the test code use Seq, with comments using Array.

I suppose we could use Seq and add something to the readme. Or, the test generator could modify the test case comments to use "seq" instead of "array". Or, we can just leave the test impl to use Array.

On the template or .keep. I added an empty test file BinarySearch.scala instead of .keep.

On a separate note - I am going to be unreachable until Thursday morning US time. 

Thanks for all your input!

Issue #331, opened 2017-01-31
Sync Up Existing Test Suites with x-common


abo64:
There are continuous changes of existing tests in x-common.
Is there a good way to keep track and update the Scala tests?

Just to list some recent changes:
- [x] scrabble-score
- [x] nucleotide-count
- [x] phone-number
- [x] prime-factors
- [x] etl
- [x] secret-handshake

ErikSchierboom:
Coincidentally, an issue for this topic has only recently been opened: https://github.com/exercism/x-common/issues/524 Perhaps you could chip in there?

abo64:
What do you think about the idea to directly parse the `canonical-data.json` file as suggested in [this Haskell issue](https://github.com/exercism/xhaskell/pull/487)?

ricemery:
I think we should generate the tests programmatically. Some of the test
suites are already built in this way.
Take a look in the testgen directory and
https://github.com/exercism/xscala/blob/master/README.md . The testgen code
I have written isn't all that clean, but it works.

abo64:
@ricemery Yes, that looks like the right idea.

I wonder whether we should even bother to generate extra Scala source code from the `TestCase`s as it complicates matters quite a bit.
Maybe just use the `TestCase`s to run the tests immediately? (very similar to the Haskell PR).

(btw, you use `play.api.libs.json`. Why?)

ricemery:
If I understand the Haskell PR correctly, they want their test code to read the JSON and build/execute test cases dynamically.
I'm not sure I like that idea. It add complexity for the end user trying to figure out how to code their solution to meet the test requirements. When I solve exercism problems, I always look at the test code to figure out how to proceed. I suppose this is TDD.
I think having the test cases obfuscated within the JSON makes doing TDD more difficult.

Though I do agree it would make maintaining the test suites easier.

I am using play json because I knew how to use the api. It is in no way a statement on play json being the best way to parse json.

abo64:
True, this approach works only if a closer look at the test suite is not required (which is the case in the Haskell track). Some issues:
- Every exercise includes a stub with the signature for the solution;
or if we don't want this we need a unified format of the tests that makes it very clear how the solution signature must be: For example always having a (possibly overloaded) function named `callMUT` or so that then calls and reveals the required signature(s).
- The test failure messages must be self-explanatory including input, expected and actual output.
- There are no `pending` exercises (which I personally would welcome - one usually tries to go through the tests from top to bottom anyway to make them green).
[...]

PR #334, opened 2017-02-06
[WIP] Add TestBuilder

abo64:
This is a first attempt on issue #331 

There is a new class `TestBuilder` (a DSL would probably be more Scala-like than the builder pattern, but the latter was straightforward and less time-consuming).

To use it one has to provide an `implicit def` to translate a test case into a `TestCaseGen`.
An example would be `AllYourBaseTestGenerator.testCaseToGen`.

Please let me know what you think.

(perhaps something similar can be done at the json parser side?)

### DIFF
--- a/testgen/src/main/scala/AllYourBaseTestGenerator.scala
+++ b/testgen/src/main/scala/AllYourBaseTestGenerator.scala
@@ -2,35 +2,28 @@ import play.api.libs.json.Json
 
 import scala.io.Source
 
+
 class AllYourBaseTestGenerator {
-  implicit val testCaseReader = Json.reads[TestCase]
+  implicit val testCaseReader = Json.reads[AllYourBaseTestCase]
 
   private val filename = "all-your-base.json"
   private val fileContents = Source.fromFile(filename).getLines.mkString
   private val json = Json.parse(fileContents)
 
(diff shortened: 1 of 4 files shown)


ricemery:
Thanks @abo64 .

This looks pretty cool. I do notice though that the TestBuilder will not work with the Bowling exercise or CustomSet.

For Bowling the test asserts are not in the form - "result should be ....". A matcher is being used so that a good error message can be returned for failures.

For CustomSet there are multiple functions being tested. I don't see a way to support multiple different TestCaseGens. Note that  CustomSet has 9 different TestCase case classes. The different case classes support different function calls and different function params/json definition.

Is there a way to handle these exercises with what you have done?

Thanks!! Again, I am going to be unreachable until Thursday. So, please don't take silence as disapproval:) !

abo64:
@ricemery I modified `TestBuilder` to be more flexible:
- Now you have to add individual groups of tests with an optional description. This should cover the case of multiple different functions to test like `CustomSet`, among other things.
- You can modify the simple default behavior by overriding `result` and `checkResult` in case class `TestCaseGen`. This should cover `Bowling`.

Not overly elegant. But perhaps it can be generalized and beautified some more?

PR #336, opened 2017-02-06
Add strain exercise

ricemery:
Add strain exercise.

There was no canonical test data. Tests are based on the C# tests.

------
Diff from last pull request.
Added pending to test cases.
Fixed example.scala to include an example impl. And, fixed Strain.scala to be empty.

@ErikSchierboom  - the formatting of config.json seems consistent with what is already there. At least visually in my browser. I remember you checking in a cleaned up config.json, but I can't seem to find that commit in the history.
What tool did you use to reformat the json? Maybe we should all use the same formatter/set of rules when changing the file.

### DIFF
--- a/exercises/strain/example.scala
+++ b/exercises/strain/example.scala
@@ -1,7 +1,22 @@
-object Strain {
-  def keep[A](traversable: Traversable[A], func: A => Boolean): Traversable[A]
-    = for (a <- traversable if func(a)) yield a
+import scala.annotation.tailrec
 
-  def discard[A](traversable: Traversable[A], func: A => Boolean): Traversable[A]
-    = for (a <- traversable if !func(a)) yield a
+object Strain {
+  def keep[A](seq: Seq[A], func: A => Boolean): Seq[A] =
+    recurse(seq, func, List()).reverse
+
+  def discard[A](seq: Seq[A], func: A => Boolean): Seq[A] = {
+    val notFunc: A => Boolean = a => !func(a)
+    recurse(seq, notFunc, List()).reverse
+  }
+
+  @tailrec
+  private def recurse[A](seq: Seq[A], func: A => Boolean, acc: Seq[A]): Seq[A] =
+    seq match {
+      case Nil => acc
+      case x::xs =>
+        if (func(x))
+          recurse(xs, func, x +: acc)
+        else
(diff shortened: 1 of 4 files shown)


exercises/strain/example.scala
```diff
@@ -0,0 +1,7 @@
+object Strain {
+  def keep[A](traversable: Traversable[A], func: A => Boolean): Traversable[A]
+    = for (a <- traversable if func(a)) yield a
```
abo64: One could argue that a for-comprehension is just syntactic sugar for `filter` and therefore violates the Readme restrictions?! :)

ErikSchierboom: I agree with @abo64, this does seem a bit like cheating :) What I think the exercise leads you to is to use a plain loop for imperative languages and recursion for functional languages.

ricemery:
@abo64 / @ErikSchierboom  - I have changed the impl to remove the for comprehension. As part of that, I simplified the tests to only test Seq and List. I removed the test cases using Set.

And, changed the example impl to take a Seq instead of a Traversable. Using Traversable with the for comprehension wasn't difficult. But, without the for comprehension, I think it makes requiring Traversable more difficult than necessary for exercise.

exercises/strain/src/test/scala/StrainTest.scala
```diff
@@ -0,0 +1,78 @@
+import org.scalatest.{Matchers, FlatSpec}
+
+class StrainTest extends FlatSpec with Matchers {
+
+  it should "handle empty keep" in {
+    val result = Strain.keep[Int](List(), x => x < 10)
+    result should be(List.empty)
```
abo64: Have a space character after each `be`?

ErikSchierboom:
@ricemery It must be some spaces/tabs problem:

image

 I've used some online JSON formatter, I think it was [this one](https://jsonformatter.curiousconcept.com/). Note that I changed the settings to use 2 spaces indentation.

ricemery:
Added a space between the test "be" statement and the open paren.
And ran config.json through a formatter - https://jsonformatter.curiousconcept.com/ .

ErikSchierboom:
The config looks good!

abo64:
thanks @ricemery

PR #337, opened 2017-02-14
Binary search

ricemery:
Update BinarySearchTest based on code review.
 

- Removed pending from first test case. 
- Removed unnecessary [Int] from Array creation.
- Made use of Optional more clear. 
- Also changed to using TestBuilder within BinarySearchTestGenerator.

### DIFF
--- a/config.json
+++ b/config.json
@@ -219,6 +219,14 @@
       ]
 		},
     {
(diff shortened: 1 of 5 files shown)


abo64:
thanks @ricemery

PR #338, opened 2017-02-18
Add additional scala track documentation

ricemery:
* Remove dead link from RESOURCES.md
* Add books and online courses to LEARNING.md
* Add some additional detail to TESTS.md

Refs #248

### DIFF
--- a/config.json
+++ b/config.json
@@ -219,6 +219,14 @@
       ]
 		},
     {
+      "slug": "binary-search",
+      "difficulty": 1,
+      "topics": [
+         "Searching",
+         "Arrays"
+      ]
+		},
+    {
       "slug": "prime-factors",
       "difficulty": 1,
       "topics": [
--- a/config.json
+++ b/config.json
@@ -1,675 +1,676 @@
 {
-  "slug": "scala",
-  "language": "Scala",
-  "repository": "https://github.com/exercism/xscala",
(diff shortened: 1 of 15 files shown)


ricemery:
Note that the book links are too Amazon. I wasn't sure where to link to. The C# track was linking to Amazon - so I just went with that.

ErikSchierboom:
Yeah, I just went with Amazon because most people know them as a relatively reliable seller, and they can just order the books quickly if they want.

abo64:
What I usually do in order not to prefer an individual reseller is to give the link of the publisher, for example:
- [Dean Wampler, Alex Payne: Programming Scala, 2nd Edition](http://shop.oreilly.com/product/0636920033073.do)
- [Paul Chiusano, Rúnar Bjarnason: Functional Programming in Scala](https://www.manning.com/books/functional-programming-in-scala)

But it seems there is no official policy about that for `exercism.io`?  
If you guys think that Amazon links are fine I won't resist.

ricemery:
I changed the book URLS to link to the publisher sites. I wasn't overly jazzed about including US Amazon links to the books. I suppose it doesn't matter that much. The most important thing is the title/author.

Also, I added a link for "Scala for the Impatient".

abo64:
thanks @ricemery

PR #339, opened 2017-02-19
Add rail-fence-cipher exercise.

ricemery:


exercises/rail-fence-cipher/example.scala
```diff
@@ -0,0 +1,46 @@
+case class RailFenceCipher(rails: Int) {
```
abo64: Why did you introduce this extra `case class`?

ricemery: Because I wanted to hold the size value as a member variable. 
I suppose I could have just had size as a variable within encode and decode and then passed the value to the track function. But, I didn't think about it.

abo64:
thanks @ricemery

PR #340, opened 2017-02-20
Add isogram exercise

ricemery:
Add isogram exercise

### DIFF
--- a/exercises/isogram/example.scala
+++ b/exercises/isogram/example.scala
@@ -1,0 +1,19 @@
+import scala.collection.mutable
+
+object Isogram {
+  def isIsogram(s: String): Boolean = {
+    val seen = mutable.HashSet[Char]()
+
+    s.filter(_.isLetter)
+      .map(_.toLower)
+      .foreach { c =>
+        if (seen.contains(c)) {
+          return false
+        } else {
+          seen.add(c)
+        }
+      }
+
+    true
+  }
+}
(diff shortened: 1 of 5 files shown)


exercises/isogram/example.scala
```diff
@@ -0,0 +1,19 @@
+import scala.collection.mutable
+
+object Isogram {
+  def isIsogram(s: String): Boolean = {
+    val seen = mutable.HashSet[Char]()
```
abo64: Shall we prefer an FP solution?  
I always advice against using mutable objects and `return` if there is an alternative FP solution, so I am forced by my inner Scala-self to do it here, too. :)

And now that I have a clean conscience: I know that the example solution is not essential. You may just forget about this comment after all. :)

ricemery:
Thanks @abo64 .

I have submitted, what I think, is a cleaner example.

exercises/isogram/example.scala
```diff
@@ -0,0 +1,8 @@
+object Isogram {
+  def isIsogram(s: String): Boolean = {
+    !s.filter(_.isLetter)
+      .map(_.toLower)
+      .groupBy(c => c)
```
abo64: Use `identity`?

exercises/isogram/example.scala
```diff
@@ -0,0 +1,8 @@
...
+      .values.exists(xs => xs.length > 1)
```
abo64: Apply eta-reduction?

ricemery:
Thanks @abo64 ! I have implemented your suggestions.

abo64:
thanks @ricemery

PR #341, opened 2017-02-20
Add perfect-numbers exercise

ricemery:
Add perfect-numbers exercise


exercises/perfect-numbers/src/test/scala/PerfectNumbersTest.scala
```diff
@@ -0,0 +1,47 @@
+import org.scalatest.{FlatSpec, Matchers}
+
+class PerfectNumbersTest extends FlatSpec with Matchers {
```
abo64: Does it make sense to add the tests to `x-common/exercises/perfect-numbers` first?

exercises/perfect-numbers/example.scala
```diff
@@ -0,0 +1,26 @@
+import NumberType.NumberType
+
+object PerfectNumbers {
+  def classify(n: Int): NumberType = {
+    var sumOfFactors = 0
```
abo64: Again, my Scala-conscience forces me to recommend an FP solution instead. :)

And again, you may just forget about this comment. :)

exercises/perfect-numbers/src/test/scala/PerfectNumbersTest.scala
```diff
@@ -0,0 +1,47 @@
...
+  it should "handle deficient - 3" in {
+    PerfectNumbers.classify(3) should be (NumberType.Deficent)
```
abo64: Typo: `Deficent` -> `Deficient`.

exercises/perfect-numbers/src/test/scala/PerfectNumbersTest.scala
ricemery: It would be nice if there was a canonical-data.json for the exercise. But, there are already 14 languages that already implement this exercise. I don't think we should wait on approval of the canonical data. Though, we could create the file and push up a pull request. And, then update our test suite when there is an "approved" version.

exercises/perfect-numbers/src/test/scala/PerfectNumbersTest.scala
ricemery: Good catch!

exercises/perfect-numbers/example.scala
ricemery: fixed!

ricemery:
@abo64  - I pushed up some changes.

* The example was modified to be more functional
* I change the test suite to be property based. I have never written a ScalaCheck property based test, so please provide any feedback.

Thanks

exercises/perfect-numbers/example.scala
```diff
@@ -0,0 +1,24 @@
+import NumberType.NumberType
+
+object PerfectNumbers {
+  def classify(n: Int): NumberType = {
+    val sumOfFactors
+    = Range(1, n)
```
abo64: Use `(1 until n)`?

exercises/perfect-numbers/example.scala
```diff
@@ -0,0 +1,24 @@
+import NumberType.NumberType
+
+object PerfectNumbers {
+  def classify(n: Int): NumberType = {
```
abo64: fully FP!
+1 :)

exercises/perfect-numbers/src/test/scala/PerfectNumbersTest.scala
```diff
@@ -0,0 +1,29 @@
+import org.scalacheck.Gen
+import org.scalatest.prop.PropertyChecks
+import org.scalatest.{Matchers, PropSpec}
+
+class PerfectNumbersTest extends PropSpec with Matchers with PropertyChecks {
+
+  property("deficient numbers") {
+    val deficientVals = Gen.oneOf(3, 7, 15, 33550337)
```
abo64: This is not a good case for Property-based Testing (PPT). Better keep the conventional tests that you had before.

For PPT you would define general "properties" (or "laws") that always hold. This is called [Generator-driven property checks](http://www.scalatest.org/user_guide/generator_driven_property_checks) in ScalaTest terminology.

One example that comes to my mind: "For all `Int`s: a number is always either deficient, perfect or abundant".
You could add this test to the conventional tests if you want.

ricemery:
Thanks @abo64 .
Implemented last set of suggestions..

abo64:
thanks @ricemery

PR #342, opened 2017-02-21
Add simple linked list

ricemery:
Add simple-linked-list exercise.

### DIFF
--- a/exercises/simple-linked-list/example.scala
+++ b/exercises/simple-linked-list/example.scala
@@ -22,7 +22,7 @@ class SimpleLinkedListImpl[T](first: Node[T], last: Node[T]) extends SimpleLinke
     case node: NonEmpty[T] => node.value
   }
 
-  def add(item: T): SimpleLinkedList[T] = synchronized {
+  def add(item: T): SimpleLinkedList[T] = {
     val newLast = NonEmpty[T](item, Empty)
 
     last match {
@@ -39,7 +39,7 @@ class SimpleLinkedListImpl[T](first: Node[T], last: Node[T]) extends SimpleLinke
     case node: NonEmpty[T] => new SimpleLinkedListImpl[T](first.asInstanceOf[NonEmpty[T]].next, last)
   }
 
-  def reverse: SimpleLinkedList[T] = synchronized {
+  def reverse: SimpleLinkedList[T] = {
     var newLast: Node[T] = Empty
     var prev: Node[T] = Empty
     var current = first
--- a/exercises/simple-linked-list/example.scala
+++ b/exercises/simple-linked-list/example.scala
@@ -1,0 +1,94 @@
+import scala.reflect.ClassTag
+
+class Node[+T]
+case object Empty extends Node[Nothing]
+case class NonEmpty[T](value: T, var next: Node[T]) extends Node[T]
+
+trait SimpleLinkedList[T] {
+  def isEmpty: Boolean
+  def value: T
+  def add(item: T): SimpleLinkedList[T]
+  def next: SimpleLinkedList[T]
+  def reverse: SimpleLinkedList[T]
+  def toArray[U >: T](implicit m: ClassTag[U]): Array[U]
+}
+
+class SimpleLinkedListImpl[T](first: Node[T], last: Node[T]) extends SimpleLinkedList[T] {
+
+  def isEmpty: Boolean = first == Empty
+
+  def value: T = first match {
+    case Empty => throw new NoSuchElementException
+    case node: NonEmpty[T] => node.value
+  }
+
+  def add(item: T): SimpleLinkedList[T] = synchronized {
+    val newLast = NonEmpty[T](item, Empty)
+
+    last match {
+      case Empty =>
+        new SimpleLinkedListImpl[T](newLast, newLast)
+      case node: NonEmpty[T] =>
+        node.next = newLast
+        new SimpleLinkedListImpl[T](first, newLast)
+    }
+  }
+
+  def next: SimpleLinkedList[T] = first match {
+    case Empty => this
+    case node: NonEmpty[T] => new SimpleLinkedListImpl[T](first.asInstanceOf[NonEmpty[T]].next, last)
+  }
+
+  def reverse: SimpleLinkedList[T] = synchronized {
+    var newLast: Node[T] = Empty
+    var prev: Node[T] = Empty
+    var current = first
+    var next: Node[T] = Empty
+    while (current != Empty) {
+      val newNode = NonEmpty[T](current.asInstanceOf[NonEmpty[T]].value, prev)
+
+      if (newLast == Empty) {
+        newLast = newNode
+      }
+
+      next = current.asInstanceOf[NonEmpty[T]].next
+      prev = newNode
+      current = next
+    }
+
+    new SimpleLinkedListImpl[T](prev, newLast)
+  }
+
+  def length: Int = {
+    var len = 0
+
+    var current = first
+    while (current != Empty) {
+      len = len + 1
+      current = current.asInstanceOf[NonEmpty[T]].next
+    }
+
+    len
+  }
+
+  def toArray[U >: T](implicit m: ClassTag[U]): Array[U] = {
+    val arr = new Array[U](length)
+
+    var current = first
+    var i = 0
+    while (current != Empty) {
+      arr(i) = current.asInstanceOf[NonEmpty[T]].value
+      current = current.asInstanceOf[NonEmpty[T]].next
+      i = i + 1
+    }
+
+    arr
+  }
+}
+
+object SimpleLinkedList {
+  def apply[T](): SimpleLinkedList[T] = new SimpleLinkedListImpl[T](Empty, Empty)
+
+  def apply[T](arr: Array[T]): SimpleLinkedList[T] =
+    arr.foldLeft(SimpleLinkedList[T]())((acc, t) => acc.add(t))
+}
--- a/exercises/simple-linked-list/example.scala
+++ b/exercises/simple-linked-list/example.scala
@@ -1,5 +1,3 @@
-import scala.reflect.ClassTag
-
 class Node[+T]
 case object Empty extends Node[Nothing]
 case class NonEmpty[T](value: T, next: Node[T]) extends Node[T]
@@ -10,7 +8,7 @@ trait SimpleLinkedList[T] {
   def add(item: T): SimpleLinkedList[T]
   def next: SimpleLinkedList[T]
   def reverse: SimpleLinkedList[T]
-  def toArray(implicit m: ClassTag[T]): Array[T]
+  def toSeq: Seq[T]
 }
 
 class SimpleLinkedListImpl[T](first: Node[T]) extends SimpleLinkedList[T] {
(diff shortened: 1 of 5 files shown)


exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
```diff
@@ -0,0 +1,56 @@
+import org.scalatest.{FlatSpec, Matchers}
+
+class SimpleLinkedListTest extends FlatSpec with Matchers {
```
abo64: I know it may be not easy for this exercise, but maybe one can agree on the tests in `x-common/exercises/simple-linked-list`?

Perhaps there are also some more useful tests in the [Haskell TestSuite](https://github.com/exercism/xhaskell/blob/master/exercises/simple-linked-list/test/Tests.hs)?

exercises/simple-linked-list/example.scala
```diff
@@ -0,0 +1,94 @@
+import scala.reflect.ClassTag
+
+class Node[+T]
+case object Empty extends Node[Nothing]
+case class NonEmpty[T](value: T, var next: Node[T]) extends Node[T]
+
+trait SimpleLinkedList[T] {
```
abo64: Maybe we should already provide this `trait` in a file `SimpleLinkedList.scala`?

exercises/simple-linked-list/example.scala
```diff
@@ -0,0 +1,94 @@
...
+  def isEmpty: Boolean
+  def value: T
+  def add(item: T): SimpleLinkedList[T]
+  def next: SimpleLinkedList[T]
+  def reverse: SimpleLinkedList[T]
+  def toArray[U >: T](implicit m: ClassTag[U]): Array[U]
```
abo64: Do you really need this `ClassTag`?  
Actually I would even demand that the element type remain the same?!

Why not just have `def toArray[T]`?

exercises/simple-linked-list/example.scala
```diff
@@ -0,0 +1,94 @@
+import scala.reflect.ClassTag
+
+class Node[+T]
+case object Empty extends Node[Nothing]
+case class NonEmpty[T](value: T, var next: Node[T]) extends Node[T]
```
abo64: And yet again for my clean conscience: Use an immutable ADT and no side effects anywhere?

And again, you can ignore this comment. :)

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  it should "handle single item list" in {
+    val list = SimpleLinkedList().add(1)
+    list.isEmpty should be (false)
+    list.value should be (1)
+  }
+
+  it should "handle single item has no next item" in {
```
abo64: Just as a remark:  
This would also be a perfect case for [Property-based testing](http://www.scalatest.org/user_guide/property_based_testing).

But it might be too demanding for most people I guess. And it is not used anywhere.

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
```diff
@@ -0,0 +1,56 @@
...
+    pending
+    val list = SimpleLinkedList().add(1)
+    list.next.isEmpty should be (true)
+  }
+
+  it should "handle two item list" in {
+    pending
+    val list = SimpleLinkedList().add(1).add(2)
+    list.value should be (1)
+    list.next.value should be(2)
```
abo64: Have a blank space consistently after every `be`?

exercises/simple-linked-list/example.scala
ricemery: toArray[T] does not make the compiler happy...
Here is some background -
* http://stackoverflow.com/questions/6085085/why-cant-i-create-an-array-of-generic-type
* https://www.chrisstucchio.com/blog/2014/cannot_find_classtag_for_element_type_T.html

I can simplify the function though..

exercises/simple-linked-list/example.scala
ricemery: Good idea.

exercises/simple-linked-list/example.scala
ricemery: Agreed. I meant to get rid of the mutability, and then forgot.. :)

exercises/simple-linked-list/example.scala
ricemery: Thinking about ClassTag more, if there isn't another solution. This might be something we should comment on in HINTS.md

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
ricemery: I'll take a look at the Haskell tests.
I started with the C# tests as my template.

Another opportunity to add a canonical-data.json :)

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
ricemery: I think we could give property based testing a shot.. At the very least it might give some exercise exercise implementers exposure to another way of testing. They may/may not like property based testing. But, the implementers just need to run the tests - not implement them.

Will property based tests be difficult to generate from the canonical data?

abo64: We would need a newly defined format for the canonical data to state those "properties" or "laws" and have them automatically generated.
But I think it could be worth the effort as there are a couple of languages with property-based testing libraries.

I have made up some properties for this exercise in the Haskell track, see [here](https://github.com/abo64/xhaskell/blob/cfc41756a0b4e9b742e08d27c4c107434f549934/exercises/simple-linked-list/test/Tests.hs#L91). We could take those and translate them to ScalaTest/ScalaCheck.

abo64:
Just played around a bit. Here is a code snippet to start with property-based testing this exercise:
```scala
  private implicit def arbitrarySimpleLinkedList[T](
      implicit arbitraryTs: Arbitrary[Seq[T]]): Arbitrary[SimpleLinkedList[T]] =
        Arbitrary {
            arbitraryTs.arbitrary map (SimpleLinkedList(_:_*))
        }

  // starting out with an arbitrary Array
  it should "arbitrary reverse 1" in {
    forAll { xs: Array[Int] =>
      assert(SimpleLinkedList(xs).reverse.toArray == xs.reverse)
    }
  }

  // starting out with an arbitrary SimpleLinkedList
  it should "arbitrary reverse 2" in {
    forAll { xs: SimpleLinkedList[Int] =>
      assert(xs.reverse.toArray == xs.toArray.reverse)
    }
  }
```

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
abo64: Just noticed: Didn't we switch to `FunSuite`?

exercises/simple-linked-list/src/main/scala/SimpleLinkedList.scala
```diff
@@ -0,0 +1,10 @@
+import scala.reflect.ClassTag
+
+trait SimpleLinkedList[T] {
+  def isEmpty: Boolean
+  def value: T
+  def add(item: T): SimpleLinkedList[T]
+  def next: SimpleLinkedList[T]
+  def reverse: SimpleLinkedList[T]
+  def toArray(implicit m: ClassTag[T]): Array[T]
```
ricemery: How do you feel about the README for the exercise specifically taking about Arrays?

From the README - 
_As a first take, lets create a singly linked list to contain the range (1..10), and provide functions to reverse a linked list and convert to and from arrays._

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
ricemery: Some of the tests are FunSuite some are FlatSpec. There really isn't any consistency, both are used.

ricemery:
@abo64  How is the following ?

```
import org.scalacheck.{Arbitrary, Gen}
import org.scalatest.prop.GeneratorDrivenPropertyChecks
import org.scalatest.{FlatSpec, Matchers}


class SimpleLinkedListTest extends FlatSpec with Matchers with GeneratorDrivenPropertyChecks {

  private implicit def arbitrarySimpleLinkedList[T](implicit arbitraryTs: Arbitrary[Array[T]]): Arbitrary[SimpleLinkedList[T]] =
    Arbitrary {
      arbitraryTs.arbitrary map (SimpleLinkedList(_))
    }

  it should "handle fromArray toArray" in {
    forAll { xs: Array[Int] =>
      assert(SimpleLinkedList(xs).toArray sameElements xs)
    }
  }

  it should "handle reverse" in {
    forAll { xs: Array[Int] =>
      assert(SimpleLinkedList(xs).reverse.toArray sameElements xs.reverse)
    }
  }

  it should "reverse back to original" in {
    forAll { xs: SimpleLinkedList[Int] =>
      assert(xs.reverse.reverse.toArray sameElements xs.toArray)
    }
  }

  it should "return correct value at index" in {
    forAll { xs: Array[Int]  =>
      whenever(xs.length > 0) {
        val simpleLinkedList = SimpleLinkedList(xs)
        val i = scala.util.Random.nextInt(xs.length)
        val value = (0 until i).foldLeft(simpleLinkedList)((acc, j) => acc.next).value
        assert(value == xs(i))
      }
    }
  }

  it should "handle generics" in {
    forAll { xs: Array[String] =>
      assert(SimpleLinkedList(xs).toArray sameElements xs)
    }
  }
}
```

exercises/simple-linked-list/src/main/scala/SimpleLinkedList.scala
abo64: Yes, I am aware of this. It is just that Scala's `Array` inherits all those bad things from Java including the need to use `ClassTag`.
But ok, if you think we should take the Readme literally then so be it. :)

abo64:
Your property-based tests look good!
Perhaps keep your earlier unit tests in addition, too?

About the test `"return correct value at index"`: How about having an access function defined similar to  [nthDatum](https://github.com/abo64/xhaskell/blob/cfc41756a0b4e9b742e08d27c4c107434f549934/exercises/simple-linked-list/test/Tests.hs#L23) and then checking all values for equality as in [this test](https://github.com/abo64/xhaskell/blob/cfc41756a0b4e9b742e08d27c4c107434f549934/exercises/simple-linked-list/test/Tests.hs#L107)?

exercises/simple-linked-list/src/main/scala/SimpleLinkedList.scala
ErikSchierboom: @ricemery I think so. In the end, you could see array as an implementation detail. Why shouldn't it be a list or sequence? IMHO, each track should be able to decide for themselves which type is useful. 

There is also precendence to not using an array, as the Haskell and F# tracks only have a `toList` method (see [here](https://github.com/exercism/xhaskell/blob/master/exercises/simple-linked-list/examples/success-standard/src/LinkedList.hs) and [here](https://github.com/exercism/xfsharp/blob/master/exercises/simple-linked-list/Example.fs)).

ErikSchierboom:
I also really like having the property-based tests. Maybe, as @abo64 suggested, you could also keep some of the (simple) value-based tests?

exercises/simple-linked-list/src/main/scala/SimpleLinkedList.scala
ricemery: Cool. Thanks.

exercises/simple-linked-list/src/main/scala/SimpleLinkedList.scala
```diff
@@ -0,0 +1,10 @@
+import scala.reflect.ClassTag
+
+trait SimpleLinkedList[T] {
```
ricemery: I was kinda assuming that the implementers could come up with the companion object themselves by the time they get to this exercise.
Do you think this is too much to expect?

ricemery:
Hey @abo64 , @ErikSchierboom ,

I just pushed up changes based on all your feedback. Let me know if I have missed anything. Or, you notice other issues.

Thx,
Ric

exercises/simple-linked-list/src/test/scala/SimpleLinkedListTest.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  it should "handle two item list has no next value" in {
+    pending
+    val list = SimpleLinkedList().add(1).add(2)
+    list.next.next.isEmpty should be (true)
+  }
+
+  it should "allow creation from a Seq" in {
+    pending
+    val list = SimpleLinkedList.fromSeq(List(3, 2, 1))
```
abo64: I think `fromSeq` should be added in the template's companion object? After all it is then part of the public API like the `trait`'s functions.

Another question is whether we should also keep the constructors in the tests or only have `fromSeq`.
The constructors make the exercise more interesting, but also somewhat harder.
A compromise could be to have the `apply` method in the template and have it calling `fromSeq`.
(see also my comment about this topic a couple of days ago).

ricemery:
Thanks Guys !

PR #343, opened 2017-02-23
Add tests to account for whitespace

DesHorsley:
Currently only the decode encode combination has whitespace, which doesn't reflect when the whitespace is being encoded like a regular character. The README explicitly states that whitespace should remain unchanged.

### DIFF
--- a/exercises/run-length-encoding/src/test/scala/RunLengthEncodingTests.scala
+++ b/exercises/run-length-encoding/src/test/scala/RunLengthEncodingTests.scala
@@ -51,7 +51,7 @@ class RunLengthEncodingTests extends FunSuite with Matchers {
 
   test("decode with whitespace") {
     pending
-    RunLengthEncoding.decode("12WB12W3B24WB") should be ("WWWWWWWWWWWWBWWWWWWWWWWWWBBBWWWWWWWWWWWWWWWWWWWWWWWWB")
+    RunLengthEncoding.decode("  hs2q q2w  ") should be ("  hsqq qww  ")
   }
 
   test("decode(encode(...)) combination") {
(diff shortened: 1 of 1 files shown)


abo64:
Thank you @DesHorsley A few remarks from my side up front:
- One could argue that the existing last test `"decode(encode(...)) combination"` also covers whitespaces? What do you think?
But you may have a point that explicit whitespace tests would be good.
- Your new test `"decode with whitespace"` does not include any whitespaces, though?
- It might make sense to also add those missing tests to [x-common/exercises/run-length-encoding](https://github.com/exercism/x-common/tree/master/exercises/run-length-encoding) so that all languages can profit from them? Here are the respective test data in file [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/canonical-data.json).

DesHorsley:
Thanks for pointing out the error in the second test, cut and past blunder, I'll sort it out shortly.

If I send a push request to the common tests, will the change propagate down automatically or are the scala tests hand written?

The  `"decode(encode(...)) combination"` test will pass if the whitespace is encoded and decoded like a character, when it seems like this should be left unchanged i.e. "  " should not encode to "2 ", assuming I've understood the readme.

DesHorsley:
Updated the pull request.
 Thanks for your patience, I'm fairly new to open source contributions but keen to help out.

abo64:
Contributions are always welcome, thank you @DesHorsley !

You are perfectly right. If they state it in the `description.md` then there should also be tests for it!

Would be nice if this update of TestSuites were automated, but it is not so easy to do for all languages. So the general answer is no.

But for some exercises we already have our own [code generators](https://github.com/exercism/xscala/tree/master/testgen/src/main/scala) to make that process easier.
If you feel up to the task you could provide one also for this exercise. Here is a [recent example](https://github.com/exercism/xscala/blob/master/testgen/src/main/scala/IsogramTestGenerator.scala) using [TestBuilder](https://github.com/exercism/xscala/blob/master/testgen/src/main/scala/TestBuilder.scala).

DesHorsley:
I'm interested in producing the TestGenerator for this exersize, I'll try look at it this weekend and see how far I get.

ricemery:
A run length encoding test generator has been created for this test. Closing as addressed.

PR #344, opened 2017-03-04
Add variable-length-quantity exercise.

ricemery:


exercises/variable-length-quantity/src/test/scala/VariableLengthQuantityTest.scala
```diff
@@ -0,0 +1,162 @@
+import org.scalatest.{FunSuite, Matchers}
+
+class VariableLengthQuantityTest extends FunSuite with Matchers {
+
+  // Encode a series of integers, producing a series of bytes.
+  test("zero") {
```
abo64: Have the function to be tested in the test descriptions? For example:
```scala
test("encode: zero") {
```

exercises/variable-length-quantity/src/test/scala/VariableLengthQuantityTest.scala
```diff
@@ -0,0 +1,162 @@
...
+    pending
+    val encoded = VariableLengthQuantity.encode(List(0x2000, 0x123456, 0xfffffff, 0x0, 0x3fff, 0x4000))
+    encoded should be (List(0xc0, 0x0, 0xc8, 0xe8, 0x56, 0xff, 0xff, 0xff, 0x7f, 0x0, 0xff, 0x7f, 0x81, 0x80, 0x0))
+  }
+
+  // Decode a series of bytes, producing a series of integers.
+  test("one byte") {
+    pending
+    val decoded = VariableLengthQuantity.decode(List(0x7f))
+    decoded should be (Right(List(0x7f)))
```
abo64: People have more difficulties with `Either` than `Option`.

So I wonder if `decode` should not simply return an `Option[List[Int]]`?

ErikSchierboom: I'm also in favor of using `Option` here.

ricemery: @abo64 , @ErikSchierboom - I don't mind changing the API. But, I wanted to return the reason for the error from the decode method. Option hides the reason.

I would think that by the time that the implementer gets to this exercise, they should be able to handle Either.

abo64: Fine with me.

But then why not also check for the concrete error message? There seem to be only two of them?

ricemery: I didn't want to force the error message content on the implementer. I was just trying to reinforce the fact that either is a good way to get the error info out.
Returning an English language string error is probably not the best way to spit out error for most APIs. So, I was more interested in `isLeft` indicating the error, and leave it to the implementer to distinguish between errors.

exercises/variable-length-quantity/src/test/scala/VariableLengthQuantityTest.scala
ricemery: Good suggestion. Thx

abo64:
thanks @ricemery 
Let's not forget to review the difficulties of these recently added exercises, too.

PR #348, opened 2017-03-27
Add a modification of non ascii chars test without character x

cogitor:
The base test seems to be modified with the non-ascii test. 

I feel that another useful test would be a string with non-ascii characters, but with a missing character 'x'. This would invalidate solutions that are written to simply pass the previous test (e.g. by making the test for distinct characters be >= 26).


ricemery:
@cogitor , 

The PangramTest is generated from the canonical data at 
- https://github.com/exercism/x-common/blob/master/exercises/pangram/canonical-data.json . Additional test cases should be added to the canonical-data. Many languages tracks also generate their test suites in a similar way to Scala. Thus updating the canonical-data is beneficial to the other language tracks.

Thanks!

cogitor:
@ricemery cool, in that case I should add both my test and the previous non-ascii one there too :)

Issue #353, opened 2017-05-01
leap hints should mention the use of case classes


zombiefungus:
This exercise is simple and in most languages the hard part would be a bunch of `if`s and some math operations, with many posible implementations. In the scala track this was solved using `case class`es but there is no way for a begginer to identify it as such, mostly because this does not exist in other languages, oposite to normal class, which you could identify as one because tests use the  `new` keyword

zombiefungus:
Also, `case class` should be part of this exercise's topics

abo64:
I agree, `leap` is indeed problematic.
But probably using a `case class` does not make much sense at all here - the exercise should rather be changed to use an object as we did already for some other exercises as in [this issue](https://github.com/exercism/xscala/issues/242).
So what do you think if we change the signature to something like
```scala
object LeapYear {
  def isLeapYear(year: Int): Boolean = ???
}
```
Feel free to create a PR.

If you want an additional challenge you can even add a test code generator [here](https://github.com/exercism/xscala/tree/master/testgen/src/main/scala) using [TestSuiteBuilder](https://github.com/exercism/xscala/blob/master/testgen/src/main/scala/testgen/TestSuiteBuilder.scala).

zombiefungus:
I agree, there is no need, but changing the signature solves the problem of not introducing case classes properly just for awhile, since there are [several exercises](https://github.com/exercism/xscala/search?utf8=%E2%9C%93&q=%22case+class%22&type=) (one of them is Space Age, the next exercise to leap) that use it, but there is not even one exercise that listes it as a topic in `config.json`. When should this concept be introduced?

abo64:
Yes, `space-age` is another exercise that could rather be a stateless object instead.

The topics are mostly taken from (not language specific) [x-common/TOPICS.txt](https://github.com/exercism/x-common/blob/master/TOPICS.txt) and AFAIK are only for internal use (for example, as a factor to determine the difficulty).
Topics should *not* be used to prescribe any solution.

zombiefungus:
So, you suggest making both `leap` and `space-age`stateless objects, and leaving case classes for latter (whatever that implies)? The I could make a PR to help

abo64:
In general the idea is to have classes or case classes only when some internal state makes good sense.

Great, your PRs are welcome! Please have one PR per exercise.

zombiefungus:
Two questions:
- what should I do about `space-age`?
- do you have any example about the test generator? (something for dummies?)
thanks

[...]

PR #354, opened 2017-05-04
Change leap to stateless (issue #353)

zombiefungus:


exercises/leap/example.scala
```diff
@@ -1,5 +1,6 @@
-case class Year(year: Int) {
-  lazy val isLeap: Boolean = divisibleBy(4) && (divisibleBy(400) || !divisibleBy(100))
+object Year {
+  def isLeap(year: Int): Boolean =
+    divisibleBy(year, 4) && (divisibleBy(year, 400) || !divisibleBy(year, 100))
 
-  private def divisibleBy(i: Int) = year % i == 0
+  private def divisibleBy(year: Int, i: Int) = year % i == 0
```
abo64: You could consider to have this as a nested function inside `isLeap` so that you can omit the `year` parameter again.

exercises/leap/src/main/scala/Leap.scala
```diff
@@ -0,0 +1,3 @@
+object Year {
```
abo64: Now this touches an interesting subject:
Some time ago we decided in issue #137 to have such a file with the expected functions for the first 10 exercises.

`leap` was originally not among those, but now after the reordering according to difficulty it is.
So I think to add this file is just fine. Am I right, @ricemery @ErikSchierboom ? Maybe we should also review the other affected exercises under this aspect?

ErikSchierboom: @abo64 As far as I can recall: yes, you are right. I think reviewing the other exercises makes sense.

exercises/leap/example.scala
zombiefungus: nice, done

zombiefungus:
well I know this is a tiny change, but there is anything else to be done in the pr? does this closses issue #353 ? (I think it suggests new issues but any way)

abo64:
Thank you @zombiefungus

PR #355, opened 2017-05-13
Add protein-translation exercise

ricemery:
Add protein-translation exercise.

Guessed at difficulty at 4.. Maybe could be lowered to 3.

Let me know if you see any problems.. 

Thx

### DIFF
--- a/exercises/protein-translation/example.scala
+++ b/exercises/protein-translation/example.scala
@@ -1,14 +1,4 @@
-/**
- * Copyright 2017 © Denmar Technical Services Inc
- * The U.S. Government has unlimited rights per DFAR 252.227-7014, all other
- * rights reserved.
- * <p>
- * WARNING - This software contains Technical Data whose export is restricted by
- * the Arms Export Control Act (Title 22, U.S.C., Sec 2751, et seq.) or the
- * Export Administration Act of 1979, as amended (Title 50, U.S. C. App. 2401
- * et seq.). Violations of these export laws are subject to severe criminal
- * penalties.
- */
+import scala.annotation.tailrec
 
 object ProteinTranslation {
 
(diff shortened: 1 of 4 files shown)


exercises/protein-translation/src/test/scala/ProteinTranslationTest.scala
```diff
@@ -0,0 +1,65 @@
+import org.scalatest.{FunSuite, Matchers}
```
abo64: Have a PR to add `canonical-data.json` first?

Although this will probably trigger a lengthy discussion. :)

ErikSchierboom: Well, there is an [open issue](https://github.com/exercism/x-common/issues/577), but as said, getting an agreement on the data could take a while. So perhaps just continue as is for the moment?

exercises/protein-translation/example.scala
```diff
@@ -0,0 +1,48 @@
...
+    "UAU" -> "Tyrosine",
+    "UAC" -> "Tyrosine",
+    "UGU" -> "Cysteine",
+    "UGC" -> "Cysteine",
+    "UGG" -> "Tryptophan",
+    "UAA" -> "STOP",
+    "UAG" -> "STOP",
+    "UGA" -> "STOP")
+
+  def translate(input: String): Option[Seq[String]] = {
```
ErikSchierboom: I'm also in favor of keeping it as simple as possible, although I believe there are some languages that also use an option.

ricemery:
The c# version throws an exception. I went with option over exception as
that was more Scala like. I haven't looked at any other languages.

ricemery:
By the way, I don't have a strong feeling one way or another. I was just
trying to be consistent with the C# version that was explicitly dealing
with invalid input

ErikSchierboom:
@ricemery Well, I'm the author of the C# version :D I think I copied it from another track, but lately I've been of the opinion that if the exercise doesn't require it, we shouldn't test for it.

ricemery:
Closing this pull. Going to open a new one.

PR #357, opened 2017-05-16
Add file with expected interface. Refs #137

ricemery:
Add file with expected interface for Bob example. Refs #137

### DIFF
--- a/exercises/bob/src/main/scala/Bob.scala
+++ b/exercises/bob/src/main/scala/Bob.scala
@@ -1,0 +1,3 @@
+class Bob {
+  def hey(statement: String): String = ???
+}


abo64:
Thanks for the PRs @ricemery ! I think they all look fine and can be merged.

But do you want to keep the `.keep` files? (sorry for the pun).

ricemery:
The .keep files can go.

But, I'll handle that as a separate PR. I'll delete the .keep files from
the first 10 exercises in 1 PR, since the lack of the .keep file will not
break anything. The other PRs I wanted to handle individually, so that each
exercise change was isolated from the others.

Thanks for the suggestion.

ErikSchierboom:
Looks great!

PR #358, opened 2017-05-16
Add file with expected interface. Refs #137

ricemery:
Add file with expected interface for SpaceAge exercise. Refs #137

Note that the example implementation uses lazy val for the different public vals. I am only exposing "val". I was thinking that exposing "lazy val" was giving too much away/forcing to much on the implementer. Though, I could be convinced otherwise.

-Ric


ErikSchierboom:
I think omitting `lazy` was definitely the right choice. We could always add a HINTS.md file to nudge people towards using a lazy solution.

PR #365, opened 2017-05-20
Remove unneeded .keep files from first 10 exercises. Refs #137

ricemery:
Remove unneeded .keep files from first 10 exercises. The .keep files are not needed since there are empty impl files in the src/main/scala dirs.


abo64:
Thank you @ricemery

PR #366, opened 2017-05-22
Implement exercise "beer-song"

matteobaglini:
Hi all, 
I have ported the "beer-song" exercise. Let me know if there are any problems.


exercises/beer-song/src/main/scala/BeerSong.scala
```diff
@@ -0,0 +1,5 @@
+class BeerSong {
```
abo64: In #137 we agreed to just have a `.keep` file in the `src/main/scala` directory (except for the first couple of exercises) so that there is some additional fun to find out the function signatures. :)

exercises/beer-song/src/test/scala/BeerSongTest.scala
```diff
@@ -0,0 +1,60 @@
+import org.scalatest._
+
+class BeerSongTest extends FunSuite {
+
+  test("first generic verse") {
+    val expected = "99 bottles of beer on the wall, 99 bottles of beer.\nTake one down and pass it around, 98 bottles of beer on the wall.\n"
+    val beerSong = new BeerSong()
+    assert(expected == beerSong.verse(99))
```
abo64: Personally I also like the `assert` macro.
Most exercises use `should be` from `org.scalatest.Matchers`.

But perhaps we `should be (more liberal)` here?  @ricemery @ErikSchierboom

exercises/beer-song/src/test/scala/BeerSongTest.scala
```diff
@@ -0,0 +1,60 @@
+import org.scalatest._
+
+class BeerSongTest extends FunSuite {
+
+  test("first generic verse") {
+    val expected = "99 bottles of beer on the wall, 99 bottles of beer.\nTake one down and pass it around, 98 bottles of beer on the wall.\n"
+    val beerSong = new BeerSong()
```
abo64: There is no internal state. So it makes more sense to have an `object` here instead of a `class`. Just:
```scala
assert(expected == Beersong.verse(99))
```

exercises/beer-song/src/test/scala/BeerSongTest.scala
```diff
@@ -0,0 +1,60 @@
+import org.scalatest._
```
abo64: Actually I have written a [Code Generator](https://github.com/exercism/xscala/blob/master/testgen/src/main/scala/BeerSongTestGenerator.scala) for this exercise, not being aware at that time that the respective exercise did not exist. :)
 
Don't know if it still works, but feel free to try it out and modify it if you find the time.

exercises/beer-song/src/test/scala/BeerSongTest.scala
```diff
@@ -0,0 +1,60 @@
...
+class BeerSongTest extends FunSuite {
+
+  test("first generic verse") {
+    val expected = "99 bottles of beer on the wall, 99 bottles of beer.\nTake one down and pass it around, 98 bottles of beer on the wall.\n"
+    val beerSong = new BeerSong()
+    assert(expected == beerSong.verse(99))
+  }
+
+  test("last generic verse") {
+    pending // remove to run test
```
abo64: The first test should *not* be `pending`.

exercises/beer-song/src/test/scala/BeerSongTest.scala
ricemery: I don't think we should be strict about about test style. I think it is good from an education standpoint to have a couple different styles used in the exercises.

ErikSchierboom: Same here. The difference is small enough that people won't have much trouble I think

exercises/beer-song/src/test/scala/BeerSongTest.scala
matteobaglini: Ok, I'll look into it, thanks.

exercises/beer-song/src/test/scala/BeerSongTest.scala
matteobaglini: Isn't _first generic verse_ the first test?

Before start coding the test suite I checked the [common](https://github.com/exercism/x-common/blob/master/exercises/beer-song/canonical-data.json) and [ruby](https://github.com/exercism/xruby/blob/master/exercises/beer-song/beer_song_test.rb) reference and it seemed to me that it was the first.

abo64:
thank you @matteobaglini

matteobaglini:
it was a pleasure. 😃

PR #369, opened 2017-06-02
Add book-store exercise

ricemery:
Add book-store exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -533,6 +533,13 @@
          ]
       },
       {
+         "slug":"book-store",
+         "difficulty":5,
+         "topics":[
+            "Recursion"
+         ]
+      },
+      {
          "slug":"nth-prime",
          "difficulty":6,
          "topics":[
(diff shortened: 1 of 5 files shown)


abo64:
@ricemery You have my approval, too. So you can just merge.

One remark:
The `targetgrouping` information in [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/book-store/canonical-data.json) is ignored.
One could soup up the exercise by letting `total` return a pair consisting of the total price and the grouping.

ricemery:
@abo64 , @ErikSchierboom  - thanks.

@abo64  - Neither the Java or C# impls use the `targetgrouping` either. Maybe @ErikSchierboom knows why? :)

ErikSchierboom:
Well, I think that's because the exercise's README states the following:

> Your mission is to write a piece of code to calculate the price of any conceivable shopping basket (containing only books of the same series), giving as big a discount as possible.

In general, I'd like exercises to only have a single goal, and I think calculating the target price is the correct one here. I assume the target grouping is just there to help people understand the solution.

ricemery:
Excellent points. Thanks

Issue #370, opened 2017-06-03
Add Version Numbers to Test Suites


abo64:
Follow-up of #331: In order to keep in sync with `x-common` we need to add the version number of the respective `canonical-data.json` to all existing test suites (N.B. in addition we also need a script to detect out-of-sync test suites). For the exact format see [HelloWorldTest.scala](https://github.com/exercism/xscala/blob/master/exercises/hello-world/src/test/scala/HelloWorldTest.scala#L3).
Preferably this could be done by writing a test suite generator using [TestSuiteBuilder](https://github.com/exercism/xscala/blob/master/testgen/src/main/scala/testgen/TestSuiteBuilder.scala), but a simple manual (update of the test suite and) addition of the version number might do for now, too.

A few exercises already have a test suite generator using `TestSuiteBuilder` but don't contain the version number, so they are still unchecked.
- [x] accumulate
- [x] acronym
- [x] all-your-base
- [x] allergies
- [x] alphametics
- [x] anagram
- [x] atbash-cipher
- [x] bank-account
- [x] beer-song
- [x] binary-search-tree
- [x] binary-search
- [ ] binary
- [x] bob
- [x] bowling
- [x] bracket-push
- [x] change
- [x] clock
- [x] connect
- [x] crypto-square
- [x] custom-set
- [x] difference-of-squares
- [x] dominoes
- [x] etl
- [x] food-chain
- [x] forth
- [x] gigasecond
- [x] grade-school
- [x] grains
- [x] hamming
- [x] hello-world
- [ ] hexadecimal
- [x] house
- [x] isogram
- [x] kindergarten-garden
- [x] largest-series-product
- [x] leap
- [x] lens-person
- [x] linked-list
- [x] luhn
- [x] matrix
- [x] meetup
- [x] minesweeper
- [x] nth-prime
- [x] nucleotide-count
- [x] ocr-numbers
- [ ] octal
- [x] palindrome-products
- [x] pangram
- [x] parallel-letter-frequency
- [x] pascals-triangle
- [x] perfect-numbers
- [x] phone-number
- [x] pig-latin
- [x] prime-factors
- [x] protein-translation
- [x] pythagorean-triplet
- [x] queen-attack
- [x] rail-fence-cipher
- [x] raindrops
- [x] rna-transcription
- [x] robot-name
- [x] robot-simulator
- [x] roman-numerals
- [x] run-length-encoding
- [x] saddle-points
- [x] say
- [x] scrabble-score
[...]

PR #371, opened 2017-06-04
Update Bob exercise to conform to canonical-data.json. Refs #370

ricemery:
* Add BobTestGenerator
* Generate BobTest.scala
* Update example.scala and Bob.scala to be objects instead of class.  The test generator defaults to object.
* Update example.scala and Bob.scala to use function name of "response" instead of "hey".
* Fix bug in example.scala to trim trailing space when verifying Question.


abo64:
thank you @ricemery

PR #375, opened 2017-06-05
Add TestGenerator for perfect-numbers. Refs #331 and #370

ricemery:
* Create PerfectNumbersTestGenerator.
* Expose TestSuiteBuilder.toString.  @abo64  - there may be a better way to expose/reuse this functionality. 
* Update PerfectNumbersTest
* Update example.scala to conform to new function signature that returns Either. 
* Update config.json to change difficulty from 2 to 3. And, move exercise farther down the ordering.


testgen/src/main/scala/PerfectNumbersTestGenerator.scala
```diff
@@ -0,0 +1,42 @@
...
+          case Left(error) => s"Left(${TestSuiteBuilder.toString(error)})"
+          case Right(exp) => s"Right(${toEnumStr(exp.toString)})"
+        }
+    }
+
+    def fromLabeledTest(argNames: String*): ToTestCaseData =
+      withLabeledTest { sut =>
+        labeledTest =>
+          val args = sutArgs(labeledTest.result, argNames: _*)
+          val property = labeledTest.property.mkString
```
abo64: Why `mkString`? `labeledTest.property` is already a string?

testgen/src/main/scala/testgen/TestSuiteBuilder.scala
```diff
@@ -86,7 +86,7 @@ object TestSuiteBuilder {
   private def toString(expected: CanonicalDataParser.Expected): String =
       expected.fold(error => s"Left(${toString(error)})", toString)
 
-  private def toString(any: Any): String = {
+  def toString(any: Any): String = {
```
abo64: I expect the builder code to evolve and adapt when more exercises are covered.
So I am fine with this change.

testgen/src/main/scala/PerfectNumbersTestGenerator.scala
ricemery: I started with a copy of PangramsTestGenerator. And, PangramsTestGenerator line 12 does the same thing. I'll remove the mkString in PerfectNumbersTestGenerator.scala  and update the pull request.

Thanks!

ricemery: BTW: I realize that the .mkString is required in PangramsTestGenerator since the value is a boolean in that case.

abo64:
thank you @ricemery

ricemery:
Thanks

PR #381, opened 2017-06-12
Create HammingTestGenerator. And, update HammingTest. Refs #370, #331

ricemery:
Create HammingTestGenerator. And, update HammingTest. Refs #370, #331


abo64:
thanks @ricemery

PR #384, opened 2017-06-17
Create EtlTestGenereator, update EtlTest and example. Refs #370, #331

ricemery:
Create EtlTestGenereator, update EtlTest and example. Refs #370, #331

* Created EtlTestGenerator. I don't see a clean way to generate "sutCalls" and "expected" values that wrap lines. So.. I ended up reformatting the generated code by hand. I think the generator should suffice for now, but can be imporved.
@abo64 : Do you have any thoughts on how to generate multiple line "sutCalls" or "exepected" vals?
* Updated EtlTest.scala
* Update example.scala to use mixed case object name.

### DIFF
--- a/exercises/etl/example.scala
+++ b/exercises/etl/example.scala
@@ -1,4 +1,4 @@
-object ETL {
+object Etl {
   def transform(m: Map[Int, Seq[String]]): Map[String, Int] = m.flatMap { case (key, strings) =>
     Map(strings.map(_.toLowerCase -> key): _*)
   }
--- a/exercises/etl/src/test/scala/EtlTest.scala
+++ b/exercises/etl/src/test/scala/EtlTest.scala
@@ -1,46 +1,32 @@
-import org.scalatest._
+import org.scalatest.{Matchers, FunSuite}
 
-class TransformTest extends FunSuite with Matchers {
-  test ("transform one value") {
-    val old = Map(1 -> Seq("WORLD"))
+/** @version 1.0.0 */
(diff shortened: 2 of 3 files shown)


abo64:
@ricemery If you want to have it all in just one line you could of course use `mkString("", ",", "\n")`.
Or write your own formatting function that inserts a newline every `n` elements or so?

ricemery:
@abo64 - I was trying to not have long lines. And, I wanted the newlines to line up cleanly with the preceding lines. Additionally I was trying to avoid having code to perform the newline indentation. 
Since the template is handling indentation for the other generators, I didn't want to have to add indentation code into the scala code for the Etl generator.

I am wondering if we want to have another template that allows for long argument lists and/or long expected values.

```
   @testCase.function(@for(argLine <- @testCase.argLines {
           @argLine 
        }) should be (@for(expectedLine <- @testCase.expectedLines {
           @expectedLine
        })
```

I'm not sure. This is just an idea that is bouncing around my head.

abo64:
@ricemery sounds good. Hardest part is probably to write such a universal formatting function.
If you can do that we could either have a second template or perhaps build it somehow into the existing one (like calling it when the arglist exceeds a certain length or so).

ErikSchierboom:
Breaking up long lists seems like a good idea. In the C# generators, I have an option to not directly pass an argument but instead introduce a (possibly multi-line) variable first.

ricemery:
I am going to merge this for now. Will readdress long arg lists as a separate pull. When I am motivated:)

PR #385, opened 2017-06-18
Create GrainsTestGenereator. Update GrainsTest. Refs #370, #331

ricemery:
Create GrainsTestGenereator. Update GrainsTest. Refs #370, #331

### DIFF
--- a/exercises/grains/src/test/scala/GrainsTest.scala
+++ b/exercises/grains/src/test/scala/GrainsTest.scala
@@ -1,52 +1,59 @@
-import org.scalatest._
+import org.scalatest.{Matchers, FunSuite}
 
+/** @version 1.0.0 */
 class GrainsTest extends FunSuite with Matchers {
-  test ("square 1") {
+
+  test("1") {
     Grains.square(1) should be (Some(1))
   }
 
-  test ("square 2") {
+  test("2") {
     pending
     Grains.square(2) should be (Some(2))
   }
 
-  test ("square 3") {
+  test("3") {
     pending
     Grains.square(3) should be (Some(4))
   }
 
-  test ("square 4") {
+  test("4") {
     pending
     Grains.square(4) should be (Some(8))
   }
 
-  test ("square 16") {
+  test("16") {
     pending
     Grains.square(16) should be (Some(32768))
   }
 
-  test ("square 32") {
-    Grains.square(32) should be (Some(2147483648L))
+  test("32") {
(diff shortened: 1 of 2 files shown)


exercises/grains/src/test/scala/GrainsTest.scala
```diff
@@ -1,52 +1,59 @@
...
+    Grains.square(0) should be (None)
+  }
+
+  test("negative square raises an exception") {
     pending
     Grains.square(-1) should be (None)
   }
 
-  test ("square bigger than 64") {
+  test("square greater than 64 raises an exception") {
```
abo64: The description `raises an exception` is inadequate here.
But this should better be fixed in `x-common`.

exercises/grains/src/test/scala/GrainsTest.scala
```diff
@@ -1,52 +1,59 @@
...
     pending
     Grains.square(65) should be (None)
   }
 
-  test ("total grains") {
+  test("returns the total number of grains on the board") {
     pending
     Grains.total should be (BigInt("18446744073709551615"))
```
abo64: When I run the generator locally I get an `IllegalStateException` in L21 instead?!

exercises/grains/src/test/scala/GrainsTest.scala
ricemery: Yes. This is a problem for 3 of the tests. The descriptions mention "Exception", but we are using None to indicate an error. I wasn't really sure what to do. I could hand edit the Test case. But, I do think this would be better addressed by proposing a change to the canonical-data.json.

exercises/grains/src/test/scala/GrainsTest.scala
ricemery: Are you saying that for 

```
    {
      "description": "returns the total number of grains on the board",
      "property": "total",
      "expected": 18446744073709551615
}
```
The generator throws an IllegalArgumentException? If this is happening than my only explanation would be that the expected value is not being parsed as a BigDecimal in your environment. Which would seem odd.

abo64: That is exactly what I am saying. But then it is probably not so much of a problem?
I will investigate some more.

abo64: Seems it is parsed as a `Double`.

PR #386, opened 2017-06-18
all-your-base: Add version number to test suite (#370)

abo64:
- Rewrite `AllYourBaseTestGenerator` to use `TestSuiteBuilder`.

**Caution**: Now all `"expected": null` test cases are translated into a `None` expectation (for example `"empty list"` test case).
Not sure if this is really what we want? If not one should rather change `x-common/canonical-data.json` instead of doing manual edits after code generation?


ErikSchierboom:
I'm all in favor of using `None`.

PR #390, opened 2017-06-19
 Create TriangleTestGenerator. Update TriangleTest. Refs #370, #331

ricemery:
*  Create TriangleTestGenerator. 
* Update TriangleTest
* Rework example.scala to conform to new class/function signature. The old example used enumeration to represent the triangle type. A triangle could only be a single type. Only being a single type is invalid according to test - "equilateral triangles are also isosceles".  I removed the enumeration from the example. Now the class supports `isosceles`, `equilateral` and `scalene` functions as spec'd in the canonical-data.json.

### DIFF
--- a/testgen/src/main/scala/TriangleTestGenerator.scala
+++ b/testgen/src/main/scala/TriangleTestGenerator.scala
@@ -1,0 +1,37 @@
+import java.io.File
+
+import testgen.TestSuiteBuilder._
+import testgen._
+
+object TriangleTestGenerator {
+  def main(args: Array[String]): Unit = {
+    val file = new File("src/main/resources/triangle.json")
+
+    def mapArgToString(arg: Any): String = {
+      arg match {
+        case vals: List[_] => vals.mkString(", ")
+        case _ => throw new IllegalStateException
+      }
+    }
+
+    def sutArgs(parseResult: CanonicalDataParser.ParseResult, argNames: String*): String =
+          argNames map (name => mapArgToString(parseResult(name))) mkString
+
+    def fromLabeledTest(argNames: String*): ToTestCaseData =
+      withLabeledTest { sut =>
+        labeledTest =>
+          val args = sutArgs(labeledTest.result, argNames: _*)
+          val func = labeledTest.property.mkString
+          val sutCall = s"""Triangle(${args}).$func"""
+          val expected =
+            labeledTest.expected.fold(Function.const("true"), x => s"$x")
+          TestCaseData(labeledTest.property + " - " + labeledTest.description, sutCall, expected)
+      }
+
+    val code = TestSuiteBuilder.build(file,
+        fromLabeledTest("sides"))
+    println(s"-------------")
+    println(code)
+    println(s"-------------")
+  }
+}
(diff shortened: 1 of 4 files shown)


testgen/src/main/scala/TriangleTestGenerator.scala
```diff
@@ -0,0 +1,37 @@
...
+      arg match {
+        case vals: List[_] => vals.mkString(", ")
+        case _ => throw new IllegalStateException
+      }
+    }
+
+    def sutArgs(parseResult: CanonicalDataParser.ParseResult, argNames: String*): String =
+          argNames map (name => mapArgToString(parseResult(name))) mkString
+
+    def fromLabeledTest(argNames: String*): ToTestCaseData =
```
abo64: Have you tried `fromLabeledTestAlt` as in [BeerSongTestGenerator](https://github.com/exercism/scala/blob/master/testgen/src/main/scala/BeerSongTestGenerator.scala#L10)? Might simplify matters a bit?

abo64: I just did a quick test with
```scala
val code = TestSuiteBuilder.build(file,
  fromLabeledTestAlt("equilateral" -> Seq("sides"), "isosceles" -> Seq("sides"),  "scalene" -> Seq("sides")))
```
and it seems to work well.

abo64: But then the output assumes a stateless object - which might be a considerable alternative? For example:
```scala
test("true if all sides are equal") { 
    Triangle.equilateral(List(2, 2, 2)) should be (true)
  }
```

ErikSchierboom: I'm fine with using a stateless object here, it's not that it requires a lot of fancy logic for which an object could be useful.

ricemery: I don't have a real strong opinion. But, I would be more comfortable with a stateless object if there was only a single method returning a type. But, in this case there are multiple methods potentially operating on the same values. And, a triangle can be multiple types..

At the end of the day - what is probably most important is what is the goal of the exercise? What is the exercise meaning to teach/re-enforce?  
It is always a bit difficult to figure out what a generic exercise is attempting to teach.

abo64: Your argument about the multiple methods makes sense to me.
So one would have to modify the generator code in in a similar way as you did.

But then we might as well just take your code. :)

ricemery: I think for now, I am going to leave the test generator and the definition as is.

I like your more simplified generator.

```
val code = TestSuiteBuilder.build(file,
  fromLabeledTestAlt("equilateral" -> Seq("sides"), "isosceles" -> Seq("sides"),  "scalene" -> Seq("sides")))
```
But... The generator I wrote adds additional info to the test descriptions that I thinks is handy.
For example. 

```
  false if no sides are equal
``` 

becomes 

```
  isosceles - false if no sides are equal
```
when my generator runs. I don't think the original description is all that helpful. This is probably better to submit new descriptions to the canonical json. But, for now having the generator add the extra info to the beginning of the comments is useful IMHO.

abo64:
thanks @ricemery

PR #392, opened 2017-07-01
Create SieveTestGenereator. Update SieveTest. Refs #370, #331

ricemery:
* Create SieveTestGenereator. 
* Update SieveTest. 
* Update example to respect new function name.

Note that this is another generator that could benefit from better support for line wrap support. I ran SieveTest through a code formatter after generation.

Refs #370, #331


abo64:
I like the idea to just use a code formatter if necessary.

PR #393, opened 2017-07-01
Create AnagramTestGenerator. Update AnagramTest. Refs #370, #331

ricemery:
* Create AnagramTestGenerator. 
* Update AnagramTest. 
* Update example to be an object instead of a class.

Should be merged after pull request #391 . The generator relies on a change to TestSuiteBuilder that is in #391.

This is another generated test I had to format before submitting.

Refs #370, #331


ErikSchierboom:
Good idea to use an object!

ricemery:
@abo64 / @ErikSchierboom :

Thanks for reviewing and merging all of the requests today.

PR #394, opened 2017-07-02
Add new exercise: spiral-matrix

abo64:


kytrinyx:
Hey @abo64 I think @ricemery or @ErikSchierboom would be in a better position than I am to review this.

abo64:
@kytrinyx I unintentionally clicked on your icon. :)
Have already removed you again as reviewer, sorry for bothering you.

kytrinyx:
Ah, no worries :)

exercises/spiral-matrix/example.scala
```diff
@@ -0,0 +1,50 @@
+import scala.annotation.tailrec
```
ricemery: Not a huge deal. But it looks like the tailrec import is not used.

abo64:
thanks for the reviews @ErikSchierboom @ricemery

ErikSchierboom:
Thank you for creating them! I'm busy creating generators for the C# track: https://github.com/ErikSchierboom/csharp/tree/master/generators Quite a bit of effort, but I think I have most use cases covered so it should be more straightforward

PR #402, opened 2017-07-27
Create PascalsTriangleTestGenerator. 

ricemery:
* Create PascalsTriangleTestGenerator. 
* Update PascalsTriangleTest. 
* Update example to conform to new function name.

Note that the test generator is written to ignore the test with `null` as the function param. I didn't think that a `null` param made much sense in the context of the defined function.
Also, for the `-1` error case, the example was returning an empty list. I left the example and test case to expect an empty list for the error case. I think there are plenty of other exercises that test `Option` and `Either`.

Ref #370, #331


ricemery:
Thanks @ErikSchierboom !

PR #406, opened 2017-07-28
Create CryptoSquareTestGenerator. Update CryptoSquareTest. Refs #370,…

ricemery:
* Create CryptoSquareTestGenerator. 
* Update CryptoSquareTest. 
* Remove unneeded empty impl.
* Rework example to conform to tests' function names.
* Fix example to work on `"Non-empty plaintext results in the combined plaintext segments` test.
* Fix example to work on `54 character plaintext results in 7 chunks, the last two padded with spaces` test.

Refs #370, #331


exercises/crypto-square/example.scala
```diff
@@ -30,10 +30,10 @@ object CryptoSquare {
       List.empty
     } else {
       val size = texts.head.length
-      // Scala's transpose must work on square Lists. So pad the List. Then
-      // strip off the padding before return.
-      val padded = texts.map(s => s.padTo(size, " "))
-      padded.transpose.map(item => item.mkString.replaceAll(" ", ""))
+      // Scala's transpose must work on Lists where all elements are same size.
+      // So pad the List. Then strip off the padding before return.
+      val padded = texts.map(s => s.padTo(size, "\u0000"))
```
ricemery: No. I am just padding the list with an unprintable char that I can easily find and remove later. There was a bug with the old impl, where space was used as a pad character. If the original string included spaces, then the removal of the pad char would also remove the originally included space(s).

I agree this is pretty sneaky. There are probably better solutions to this exercise.

ErikSchierboom: Ah okay. Thanks for explaining.

PR #410, opened 2017-08-01
Update BracketPushTestGenerator. Update BracketPushTest. Refs #370, #331

ricemery:
* Update BracketPushTestGenerator. 
* Update BracketPushTest. Rename from BracketsTest.scala.
* Update example.scala to use new class name, and function name.

Note that the BracketPushTestGenerator does not produce a valid test for `complex latex expression` . The resulting function param is not escaped properly. I will go back and fix as a separate pull request. 
For now, it was easy to fix the escaping by hand.

Refs #370, #331


ErikSchierboom:
Thanks!

PR #413, opened 2017-08-09
Update CustomSetTestGenerator.scala. Update CustomSetTest.scala. Refs #370, #331

ricemery:
* Update CustomSetTestGenerator.scala.  I am not real thrilled on how I had to modify the json paths to reference the json elements for the updated json. The elements are referenced by index. 

I am not aware of a cleaner way to reference the json elements for each function call. I thought about rewriting the generator using the new generator scheme. But, the updated json format along with how the exercise is implemented would make this difficult...

* Update CustomSetTest.scala. 

Refs #370, #331


ricemery:
Thanks!

PR #419, opened 2017-08-18
Create HouseTestGenerator.scala. Update HouseTest.scala. Refs #370, #331

ricemery:
* Create HouseTestGenerator.scala. 
* Update HouseTest.scala. 
* Update example.scala to support both `verse` and `verses` functions

Refs #370, #331


ricemery:
Thanks @ErikSchierboom

PR #420, opened 2017-08-18
Update IsogramTestGenerator.scala. Update IsogramTest.scala. Refs #37…

ricemery:
* Update IsogramTestGenerator.scala. 
* Update IsogramTest.scala. 
* Replace empty impl file with `.keep`

Refs #370, #331

### DIFF
--- a/exercises/isogram/src/test/scala/IsogramTest.scala
+++ b/exercises/isogram/src/test/scala/IsogramTest.scala
@@ -1,52 +1,49 @@
-import org.scalatest.{FunSuite, Matchers}
+import org.scalatest.{Matchers, FunSuite}
 
+/** @version 1.1.0 */
 class IsogramTest extends FunSuite with Matchers {
-
+  
   test("empty string") {
-    val result = Isogram.isIsogram("")
-    result should be (true)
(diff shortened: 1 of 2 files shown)


exercises/isogram/src/test/scala/IsogramTest.scala
```diff
@@ -1,52 +1,49 @@
...
+import org.scalatest.{Matchers, FunSuite}
 
+/** @version 1.1.0 */
 class IsogramTest extends FunSuite with Matchers {
-
+  
   test("empty string") {
-    val result = Isogram.isIsogram("")
-    result should be (true)
+    Isogram.isIsogram("") should be (true)
```
ErikSchierboom: Are the parentheses in `(true)` necessary?

ricemery:
The parens are necessary if "should be" is used. If "shouldBe" is used as
the matcher then no parens are needed.

I am not sure that one style is preferred over the other. I have just
always used "should be".

ErikSchierboom:
Ah okay. I think it's fine then!

ricemery:
Thanks!

PR #423, opened 2017-08-22
Create NthPrimeTestGenerator. Update NthPrimeTest. Refs #370, #331

ricemery:
* Create NthPrimeTestGenerator. 
* Rename PrimeTest to NthPrimeTest.
* Update NthPrimeTest.
* Rename example object impl to NthPrime. Change function name.

Refs #370, #331


ricemery:
Thanks!

PR #424, opened 2017-08-22
Create OcrNumbersTestGenerator.scala. Update OcrNumbersTest.scala. Re…

ricemery:
* Create OcrNumbersTestGenerator.scala. 
* Rename OcrTest to OcrNumbersTest.
* Update OcrNumbersTest.scala. 
* Changed example impl functiomn signature and object name.
* Changed example impl to take List[String] instead of single String.

Refs #370, #331


ricemery:
Thanks

PR #426, opened 2017-08-23
Create PigLatinTestGenerator.scala. Update PigLatinTest.scala. Refs #…

ricemery:
* Create PigLatinTestGenerator.scala. 
* Update PigLatinTest.scala. 
* Updated example.scala to pass the tests. I could not find pig latin rules that mentioned special case handling of 'xr' and 'yt' prefixes. So, I just got the tests to pass. There may be other rules that the example does not implement. The solution matches Haskell's solution...

Refs #370, #331


ricemery:
Thanks!

PR #430, opened 2017-08-31
Create SpaceAgeTestGenerator.scala. Update SpaceAgeTest.scala. 

ricemery:
* Create SpaceAgeTestGenerator.scala. 
* Update SpaceAgeTest.scala. 
* Remove example stub and replace with .keep file.
* Update example to be an Object instead of Case Class.

Refs #370, #331


ricemery:
Thanks @ErikSchierboom !

PR #433, opened 2017-09-12
Update VariableLengthQuantityGenerator.scala. Update VariableLengthQu…

ricemery:
* Update VariableLengthQuantityGenerator.scala - use new format for generators.
* Update VariableLengthQuantityTest.scala. 
* Remove stub class and replace with .keep file. 

Refs #370, #331


ricemery:
Thanks!

PR #460, opened 2017-09-27
Create PalindromeProductsTestGenerator.scala. Update PalindromeProduc…

ricemery:
* Create PalindromeProductsTestGenerator.scala. 
* Update PalindromeProductsTest.scala. Note that I am including comments in the test suite to point out how the return types are used. The return types are fairly complicated..
* Update example.scala to return Option from functions.
* Update config.json to add Optional Values to topics for PalindromeProducts.

Refs #370, #331

### DIFF
--- a/config.json
+++ b/config.json
@@ -746,6 +746,7 @@
       "unlocked_by": "sum-of-multiples",
       "difficulty": 6,
       "topics": [
+        "Optional values",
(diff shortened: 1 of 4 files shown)


ricemery:
Thanks @ErikSchierboom

PR #464, opened 2017-10-02
Add VersionCheck tool. Refs #370, #331

ricemery:
This is a simple tool to compare the version of the unit test against the canonical-data.json. If there is a version mismatch info is printed to the console.

Exercises can be excluded from checking by including the exercise name in `testgen/src/main/resources/version-check-ignore.txt`. Each excluded line should be placed on a separate line.

Note that there isn't much in the way of error checking or robustness features in the tool. And, I have only tested it on macOS. I didn't take much care in dealing with OS dir separators.

### DIFF
--- a/testgen/src/main/resources/version-check-ignore.txt
+++ b/testgen/src/main/resources/version-check-ignore.txt
@@ -1,0 +1,5 @@
+binary
+hexadecimal
+octal
+trinary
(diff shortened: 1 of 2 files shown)


ricemery:
Thanks @ErikSchierboom

PR #473, opened 2017-10-03
Cleanup Meetup exercise. Refs #11

ricemery:
* Rework API to match the C# API
* Update example and test cases.
* Add emply impl that includes Schedule and constants for API call.


ricemery:
Thanks!

PR #480, opened 2017-11-14
add flatten-array exercise

trybeee:


ricemery:
Hey @trybeee ,

Thanks for writing this exercise. It has been merged. I just ran the test code through the scalafmt'r . And, I set all the test cases to pending except the first. Having all but the first test as pending is the standard in the scala track.

On a related note, most of the scala exercises have test code generators. The generators use json definitions of the test cases to spit out the test classes. The definitions for different exercises can be found in https://github.com/exercism/problem-specifications/tree/master/exercises . Look at the canonical-data.json files. If a exercise directory does not have a canonical-data.json file, then the tests have to be generated by hand.

The idea is that when the canonical-data.json changes for an exercise, we can just rerun the appropriate generator to update the test code.

It doesn't matter that you did not write a generator. We appreciate all the help we can get. Just thought I would let you know in case you attack another exercise, and would like to also write a generator.
Note that some of the generators are really simple to create (e.g. BobTestGenerator, PigLatinTestGenerator), others are a pain .

For flatten-array I wrote the test generator and submitted a pull request. You can see it at
https://github.com/exercism/scala/pull/482

Thanks again

trybeee:
@ricemery thanks for the feedback. I'll submit test code generator next time.

PR #481, opened 2017-11-15
Change flatten-array to get unlocked by accumulate instead of null

ricemery:


ErikSchierboom:
Whoops! Thanks for fixing @ricemery!

Issue #488, opened 2018-02-05
Update exercises to match latest problem specifications


ricemery:
- [ ] acronym implemented version - 1.1.0	canonical version - 1.3.0
- [ ] all-your-base	implemented version - 1.1.0	canonical version - 2.2.0
- [ ] allergies	implemented version - 1.0.0	canonical version - 1.1.0
- [ ] alphametics	implemented version - 1.0.0	canonical version - 1.2.0
- [ ] anagram	implemented version - 1.0.1	canonical version - 1.2.0
- [ ] atbash-cipher	implemented version - 1.0.0	canonical version - 1.1.0
- [ ] beer-song	implemented version - 1.0.0	canonical version - 2.1.0
- [ ] binary-search	implemented version - 1.0.0	canonical version - 1.1.0
- [ ] binary-search-tree	implemented version - created manually	canonical version - 1.0.0
- [ ] book-store	implemented version - 1.0.1	canonical version - 1.3.0
- [ ] bracket-push	implemented version - 1.1.0	canonical version - 1.2.0
- [ ] change	implemented version - 1.1.0	canonical version - 1.2.0
- [ ] clock	implemented version - 1.0.1	canonical version - 2.2.1
- [ ] collatz-conjecture	implemented version - 1.1.1	canonical version - 1.2.0
- [ ] connect	implemented version - 1.0.0	canonical version - 1.1.0
- [ ] crypto-square	implemented version - 2.0.0	canonical version - 3.2.0
- [ ] custom-set	implemented version - 1.0.1	canonical version - 1.3.0
- [ ] difference-of-squares	implemented version - 1.1.0	canonical version - 1.2.0
- [ ] dominoes	implemented version - 2.0.0	canonical version - 2.1.0
- [ ] implemented version - 1.1.0	canonical version - 1.2.0
- [ ] food-chain	implemented version - 1.0.0	canonical version - 2.1.0
- [ ] forth	implemented version - 1.2.0	canonical version - 1.4.0
- [ ] gigasecond	implemented version - 1.0.0	canonical version - 1.1.0
- [ ] hamming	implemented version - 2.0.0	canonical version - 2.1.0
- [ ] hello-world	implemented version - 1.0.0	canonical version - 1.1.0
- [ ] house	implemented version - 1.0.0	canonical version - 2.0.0
- [ ] isogram	implemented version - 1.1.0	canonical version - 1.2.0
[...]

PR #498, opened 2018-02-14
Update book-store to match latest canonical json.

ricemery:
Update book-store to match latest canonical json.
Refs #488


ricemery:
Thanks Erik!

PR #549, opened 2018-03-14
Update sum-of-multiples exercise to match latest canonical json.

ricemery:
Update sum-of-multiples exercise to match latest canonical json.
Refs #488


ricemery:
Thanks @ErikSchierboom !

PR #565, opened 2018-04-11
Added missing case that is in tests, but not here

Algiras:
https://github.com/exercism/scala/blob/master/exercises/bob/src/test/scala/BobTest.scala#L48

### DIFF
--- a/exercises/bob/README.md
+++ b/exercises/bob/README.md
@@ -6,6 +6,8 @@ Bob answers 'Sure.' if you ask him a question.
 
 He answers 'Whoa, chill out!' if you yell at him.
 
+He answers 'Calm down, I know what I'm doing!' if you yell a question at him.
+
 He says 'Fine. Be that way!' if you address him without actually saying
 anything.
 


ErikSchierboom:
Hi @Algiras. First of all, thanks for the PR! It's really appreciated. However, this PR directly modifies the test file, whereas most exercises' test suites in the Scala track are automatically being generated. You can find the generator for the Bob exercise here: https://github.com/exercism/scala/blob/master/testgen/src/main/scala/BobTestGenerator.scala

So to update the test suite, what needs to be done is to:

1. Run the test generator for the Bob exercise (@ricemery will know how to do that).
1. Verify that the existing [example implementation](https://github.com/exercism/scala/blob/master/exercises/bob/example.scala) still passes all the tests.

ricemery:
@ErikSchierboom it looks like @Algiras just changed the readme. And, linked to the previous code change in the test code. This change looks fine to me. I am going to go ahead and merge it.

Thanks @ErikSchierboom  and @Algiras  .

PR #573, opened 2018-07-22
Update forth to latest canoncial json. Refs #488 

ricemery:
Update forth to latest canoncial json. Refs #488 

Had to update the example to pass a couple of the new unit tests. The example code was failing when a UserDef was redefined after use. Added code to  grab the original definition of a UserDef when used within another UserDef.


ricemery:
Thanks for your help Erik!

PR #577, opened 2018-07-24
Add diamond exercise

ricemery:
Add diamond exercise

### DIFF
--- a/config.json
+++ b/config.json
@@ -1064,7 +1064,7 @@
       "uuid": "e45f95de-8688-46ba-80d1-470c4072c740",
       "core": false,
       "unlocked_by": "accumulate",
-      "difficulty": 2,
+      "difficulty": 4,
       "topics": [
         "lists",
         "strings",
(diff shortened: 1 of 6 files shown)


config.json
```diff
@@ -1058,6 +1058,18 @@
       "difficulty": 0,
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "diamond",
+      "uuid": "e45f95de-8688-46ba-80d1-470c4072c740",
+      "core": false,
+      "unlocked_by": "accumulate",
+      "difficulty": 2,
```
ErikSchierboom: Is this the right difficulty? I struggled with it a bit in other languages. My educated guess would put it at a difficulty of 4 or 5, but feel free to disagree :)

ricemery: I'm not sure. These numbers are so subjective. Honestly, when I updated the config.json I was thinking the difficulties were 1-5. Now I see the range is double that.

With a renewed understanding of the difficulty number I am torn between 3 and 4. I just went ahead with 4..

Thanks for all your help!

ErikSchierboom: You're welcome!

PR #582, opened 2018-08-23
Update collatz example to be include tail recursive annotation.

ricemery:
Update collatz example to be include tail recursive annotation.


ErikSchierboom:
Nice work.

ricemery:
Thanks Erik!

Issue #583, opened 2018-08-24
The space age exercise unit tests should be reworked to allow for a small delta


ricemery:
The space age exercise unit tests should be reworked to allow for a small delta. Currently the tests are expecting an exact match.

PR #586, opened 2018-10-03
Add epsilon to SpaceAge asserts. Refs #583

ricemery:
Add epsilon to SpaceAge asserts. Refs #583


ricemery:
Thanks @ErikSchierboom

PR #589, opened 2018-11-13
Remove () from methods not returning unit on robot-simulator

mlopes:
The convention in Scala is to only use () in a function/method that
doesn't take parameters. In this case the return type is not unit, so
the example should not have parenthesis.

I think this is important as these exercises often are part of people's first contact with the language, so seeing idiomatic examples that follow conventions can help them learn.


ricemery:
Thanks @mlopes

PR #606, opened 2019-03-21
Update .keep

felixrichnau:
Whenever I try to download this file for the exercise, it's just named ".keep" and empty.


ricemery:
The keep files are there just to make sure that the directory gets created. The file is not intended to be a template for the implementation.

PR #610, opened 2019-04-25
Update book-store to version 1.4 of the canonical test cases

ricemery:
Update book-store to version 1.4 of the canonical test cases.

Updated to a different sample solution that passes the new test cases.


ricemery:
Thanks @ErikSchierboom !

PR #611, opened 2019-06-08
Add two-fer exercise

ricemery:
Add two-fer exercise.

Changed `null` test case to use empty params case instead of `null`. `null` is typically avoided in Scala. And, I did not want to introduce `Option` for an early exercise. 
This has the added benefit of introducing optional params.

### DIFF
--- a/config.json
+++ b/config.json
@@ -15,6 +15,16 @@
       ]
     },
     {
+      "slug": "two-fer",
+      "uuid": "9d56454b-410a-435f-94e9-043db8c8b790",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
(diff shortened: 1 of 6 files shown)


config.json
```diff
@@ -14,6 +14,16 @@
         "strings"
       ]
     },
+    {
+      "slug": "two-fer",
```
ErikSchierboom: Maybe it is a good idea to have this exercise also unlock one or two other exercises?

ricemery: Good point. thanks

ErikSchierboom:
Great work!

ricemery:
Thanks @ErikSchierboom !

PR #612, opened 2019-06-10
Add template solution to Twofer.

ricemery:
Add an empty solution for two-fer.
There was an intent long ago to have template solutions for the first handful of the Scala core exercises. Since two-fer is now the second exercise, after hello-world, two-fer should have a template.


ErikSchierboom:
Great!

PR #613, opened 2019-06-11
Move sum-of-multiples to after leap and bob. Remove book-store as a c…

ricemery:
Reworking the progression on the core exercises.
* Move sum-of-multiples to after leap, bob and space-age.  sum-of-multiples seems to be more difficult than the other '1' difficulty exercises.
* Remove book-store as a core exercise. This one is just a major pain to mentor. I think there are probably other exercises that could be put in the book-store slot that would be more useful. And, easier to mentor. For now, I am just removing book-store...

### DIFF
--- a/config.json
+++ b/config.json
@@ -926,7 +926,7 @@
       "slug": "connect",
       "uuid": "5ea6b0e1-513e-462f-9741-46f060638a92",
       "core": false,
-      "unlocked_by": "book-store",
+      "unlocked_by": "hamming",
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Hmmm, the build failed, but it looks like an intermittent issue. Is it possible to re-start the build?

ricemery:
@ErikSchierboom - reran the tests... Thanks!

ErikSchierboom:
Thanks!

PR #615, opened 2019-07-19
incompatible  scalatest dependency for Collatz-conjecture exercise

MohamedHajr:
Fixing dependency conflicts for exercise Collatz-conjecture that causing `ClassNotFoundException Product$class`


ricemery:
THanks @MohamedHajr !

PR #618, opened 2019-08-06
Attempt to fix Travis CI failure by using a different JDK

ricemery:
Error on travis - Expected feature release number in range of 9 to 14, but got: 8

Attempt to fix Travis CI failure by using a different JDK. Using JDK 11 instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: scala
 scala:
 - 2.11.8
 jdk:
-- oraclejdk8
+- openjdk11
 before_script:
 - bin/fetch-configlet
 - sbt test:compile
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Hmmm, it did not work: `[error] (compile:compileIncremental) java.lang.NullPointerException`. Ouch.

ricemery:
@ErikSchierboom  - JDK 11 is not really needed. So, will try Open JDK 8. 
I think that exception comes when there is a null input to Regex. I don't understand the Travis build process well enough to understand what is really going on here.

ErikSchierboom:
That fixed it! Thanks 🎉

PR #629, opened 2020-01-13
Remove duplicate stub file from Grade-School exercise

ricemery:
Remove duplicate stub file from Grade-School exercise. 

There was both a stub with example functions, and a empty stub file. The empty file was removed.


ErikSchierboom:
Good catch! Thanks @ricemery.

PR #631, opened 2020-02-11
Update maintainers.json

outkaj:


config/maintainers.json
```diff
@@ -50,6 +50,16 @@
       "link_url": "https://www.leobenkel.com/",
       "avatar_url": "https://en.gravatar.com/userimage/48531829/6fcc20129e7db2a08bd76e7b80083308.png?size=200",
       "bio": null
+    },
+    {
+      "github_username": "outkaj",
+      "alumnus": true,
+      "show_on_website": true,
```
ErikSchierboom: Could you set this to `false`, which is our current default for new maintainers.

```suggestion
      "show_on_website": false,
```

Thanks!

PR #670, opened 2021-04-14
Move example files to .meta/Example.scala

ErikSchierboom:
Currently, the example.scala file is in the root of the exercise directory. This breaks the test runner as it is being included in the compilation and there are then two conflicting classes. This PR moves the example files to .meta/Example.scala, which fixes that _and_ adheres to the spec.


iHiD (approved):

iHiD:
Nice.

Issue #708, opened 2021-11-15
Unit tests for exercise 2 "Twofer" do not compile


brownbonnie:
Compile error: Unspecified value parameters: name: String

The following line is missing en empty string:
`Twofer.twofer() should be ("One for you, one for me.")`

It should be this instead:
`Twofer.twofer("") should be ("One for you, one for me.")`

ErikSchierboom:
This is not actually a bug, but a feature. See #709.

PR #709, opened 2021-11-15
Fixed Unit Test, missing empty string for first test case

brownbonnie:
I've added a small fix to get the unit tests into a compiling state.

### DIFF
--- a/exercises/practice/two-fer/src/test/scala/TwoferTest.scala
+++ b/exercises/practice/two-fer/src/test/scala/TwoferTest.scala
@@ -4,7 +4,7 @@ import org.scalatest.{Matchers, FunSuite}
 class TwoferTest extends FunSuite with Matchers {
 
   test("no name given") {
-    Twofer.twofer() should be ("One for you, one for me.")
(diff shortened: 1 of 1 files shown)


brownbonnie:
I've raised this in an issue also: https://github.com/exercism/scala/issues/708

ErikSchierboom:
This is not actually a bug, but a feature :) The goal of this exercise is to have students use optional parameters with a default value. Hence the test without a name being passed.

brownbonnie:
Ah very interesting, thanks. Would it be possible to add examples of default parameters in the tutorial?

mrybak834:
Hello,

I would argue that a test not compiling is not a result of the failure of the one trying to solve the question.

The tests are supposed to be supportive units outside of the structure of the solution program.

I don't think it makes sense for a test to not compile like it does here. 
At the very least it should be mentioned somewhere.

ErikSchierboom:
The ergonomics here are a bit odd. It might be better to convert the exercise to use an `Option[T]` parameter instead.

PR #731, opened 2022-08-17
Improve instructions and fix tests for PyhagoreanTriplet exercise.

LarsWestergren:
Fixes https://github.com/exercism/scala/issues/636

### DIFF
--- a/exercises/practice/pythagorean-triplet/.docs/instructions.md
+++ b/exercises/practice/pythagorean-triplet/.docs/instructions.md
@@ -4,7 +4,7 @@ A Pythagorean triplet is a set of three natural numbers, {a, b, c}, for
 which,
 
 ```text
-a**2 + b**2 = c**2
+a² + b² = c²
 ```
 
 and such that,
@@ -16,7 +16,7 @@ a < b < c
 For example,
 
 ```text
-3**2 + 4**2 = 9 + 16 = 25 = 5**2.
+3² + 4² = 9 + 16 = 25 = 5².
 ```
 
 Given an input integer N, find all Pythagorean triplets for which `a + b + c = N`.
--- a/exercises/practice/pythagorean-triplet/.docs/instructions.md
+++ b/exercises/practice/pythagorean-triplet/.docs/instructions.md
@@ -4,7 +4,7 @@ A Pythagorean triplet is a set of three natural numbers, {a, b, c}, for
 which,
 
 ```text
-a² + b² = c²
+a**2 + b**2 = c**2
 ```
 
 and such that,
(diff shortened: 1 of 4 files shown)


exercises/practice/pythagorean-triplet/.docs/instructions.md
```diff
@@ -16,9 +16,15 @@ a < b < c
...
-3**2 + 4**2 = 9 + 16 = 25 = 5**2.
+3² + 4² = 9 + 16 = 25 = 5².
 ```
 
+Implement the method `isPythagorean((Int, Int, Int)): Boolean` which checks the given triplet. 
+
+Then implement `pythagoreanTriplets(min: Int, max: Int): Seq[(Int, Int, Int)]` which should returns a sequence of all
+triplets whose sides are in the range `min` to `max` inclusive
+
+Finally implement `pythagoreanTripletsSum(n: Int): Seq[(Int, Int, Int)]`
```
ErikSchierboom, line 27: Editing this file unfortunately is discouraged. That is due to the contents of the file being synced from a shared exercise source, in this case: https://github.com/exercism/problem-specifications/blob/main/exercises/pythagorean-triplet/description.md

If you need to add track-specific instructions, see https://exercism.org/docs/building/tracks/practice-exercises#h-file-docs-instructions-append-md

LarsWestergren: Ah, I see. I was a bit surprised that descriptions weren't centralized, but since the Scala docs differed...

Reverting. I hope the easier to read "3²" will be synced to the Scala repo after my commit is merged.

exercises/practice/pythagorean-triplet/src/main/scala/PythagoreanTriplet.scala
```diff
@@ -0,0 +1,5 @@
+object PythagoreanTriplet {
+  def isPythagorean(nums: (Int, Int, Int)): Boolean = ???
+  def pythagoreanTriplets(low: Int, high:Int): Seq[(Int, Int, Int)] = ???
+  def pythagoreanTripletsSum(p: Int): Seq[(Int, Int, Int)] = ???
```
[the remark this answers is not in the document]
LarsWestergren: I see. Clarification of expectations since it seemed confusing. But this feels like a breaking change. I will revert this and fix the tests instead. It seems all other languages I check do not test the ordering of the 3 arguments, so it is up to the implementer to decide if this is a requirement to pass or not. Both interpretations are certainly possible.

LarsWestergren:
Ready for a new review, sorry for all the commits.

After checking several other languages (Go, Java, F#, Ruby, Haskell) they all have tests for the method/function the problem mainly discusses - "Given an input integer N, find all Pythagorean triplets", so I added tests and an implementation for that. Having it makes it clearer what the other methods do, so instructions shouldn't need changing.

Removed the automatic sorting of inputs in .meta/Example.scala.

Should I squash all the commits into one?

exercises/practice/pythagorean-triplet/.docs/instructions.md
ErikSchierboom: You could do this yourself via configlet. First run `./bin/fetch-configlet`. Then do: `./bin/configlet sync --docs --exercise pythagorean-triplets --update --yes`. We can also do this later 🤷

ErikSchierboom (approved):

ErikSchierboom:
@LarsWestergren CI is failing. It looks to be related to the changes in this PR. Could you look into it?

LarsWestergren:
> @LarsWestergren CI is failing. It looks to be related to the changes in this PR. Could you look into it?

Embarassing, missed comitting my last change.  Fixed now, and also used configlet to sync the docs.

ErikSchierboom:
Merged! Thanks 🎉

PR #735, opened 2022-09-01
Add build.sbt to files.invalidate for all practice exercises

LarsWestergren:
Pre scala 3 upgrade, so the upgrade will prompt people to update their started exercises.

### DIFF
--- a/config.json
+++ b/config.json
@@ -30,6 +30,9 @@
     ],
     "exemplar": [
       ".meta/Exemplar.scala"
+    ],
+    "invalidator": [
+      "build.sbt"
     ]
   },
   "exercises": {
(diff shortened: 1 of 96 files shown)


ErikSchierboom (approved):
Excellent! There is one other place where you can put this: https://github.com/exercism/scala/blob/main/config.json#L21

That `files` key is used to prepopulate values using `configlet sync --filepaths`.

LarsWestergren:
> Excellent! There is one other place where you can put this: https://github.com/exercism/scala/blob/main/config.json#L21
> 
> That `files` key is used to prepopulate values using `configlet sync --filepaths`.

Added!

ErikSchierboom:
Great work! Thanks a lot.

Issue #739, opened 2022-11-16
Add Code Skeleton for SpaceAge.scala


jabertuhin:
As I was going to submit the solution for SpaceAge problem using Scala language, I found it really difficult to pass the testcases. As the skeleton class isn't provided.

[* I would like to create a PR against this issue.]

ErikSchierboom:
Yes please, a PR would be much appreciated!

PR #740, opened 2022-11-16
Add SpaceAge skeleton class.

jabertuhin:
This PR is created against #739


exercises/practice/space-age/src/main/scala/SpaceAge.scala
```diff
@@ -0,0 +1,17 @@
+object SpaceAge{
```
ErikSchierboom, line 1:
```suggestion
object SpaceAge {
```

ErikSchierboom (approved):
Neat!

ErikSchierboom:
Thanks!

PR #743, opened 2022-11-27
Add code skeleton for matching brackets.

jabertuhin:
Problem Link: [Matching Brackets](https://exercism.org/tracks/scala/exercises/matching-brackets)

I have added object **MatchingBrackets**, so that it helps the practitioner while solving the problem.


exercises/practice/matching-brackets/src/main/scala/MatchingBrackets.scala
```diff
@@ -0,0 +1,3 @@
+object MatchingBrackets {
+  def isPaired(brackets: String): Boolean = ???
```
ErikSchierboom, line 2: Nice work!

ErikSchierboom (approved):

PR #744, opened 2022-11-27
Add code skeleton for Etl exercise.

jabertuhin:
Problem Link: [ETL](https://exercism.org/tracks/scala/exercises/etl)

I have added object **Etl**, so that it helps the practitioner while solving the problem.

### DIFF
--- a/exercises/practice/etl/src/main/scala/Etl.scala
+++ b/exercises/practice/etl/src/main/scala/Etl.scala
@@ -1,4 +1,3 @@
 object Etl {
-
   def transform(scoreMap: Map[Int, Seq[String]]): Map[String, Int] = ???
 }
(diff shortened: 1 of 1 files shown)


exercises/practice/etl/src/main/scala/Etl.scala
```diff
@@ -0,0 +1,3 @@
+object Etl{
```
ErikSchierboom, line 1:
```suggestion
object Etl {

```

exercises/practice/etl/src/main/scala/Etl.scala
```diff
@@ -0,0 +1,4 @@
+object Etl {
+
```
ErikSchierboom, line 2:
```suggestion

```

ErikSchierboom (approved):
I've made two minor code formatting changes. Lovely work!

PR #751, opened 2023-02-02
Hamming approaches

bobahop:
First draft


bobahop:
@ErikSchierboom Scala Hamming approaches ready for review. Tested with Scala version 3.2.2 and sbt version 1.8,2, so it should survive an upgrade to version 3.

PR #754, opened 2023-02-03
Hamming approaches tweak

bobahop:
Added the following text to the match and recursion approach:

A more idiomatic way to do the same thing as the `substring()` method calls is to use the [`head()`][head] and [`tail()`][tail] methods.
The `head()` method returns the first `Char` in the `String`, and the `tail()` method returns all of the `Char`s after the first `Char`, like so:

```scala
distanceCount(
  (if (s1.head != s2.head) acc + 1 else acc),
  s1.tail,
  s2.tail
)
```

[head]: https://www.scala-lang.org/api/2.12.7/scala/collection/immutable/StringOps.html#head:A
[tail]: https://www.scala-lang.org/api/2.12.7/scala/collection/immutable/StringOps.html#tail:Repr

### DIFF
--- a/exercises/practice/hamming/.approaches/introduction.md
+++ b/exercises/practice/hamming/.approaches/introduction.md
@@ -62,12 +62,13 @@ object Hamming {
       case 0 => acc
       case _ =>
         distanceCount(
-          (if (s1.charAt(0) != s2.charAt(0)) acc + 1 else acc),
-          s1.substring(1),
-          s2.substring(1)
+          (if (s1.head != s2.head) acc + 1 else acc),
+          s1.tail,
+          s2.tail
         )
     }
   }
+
 }
 ```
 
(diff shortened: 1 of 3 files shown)


bobahop:
@ErikSchierboom: a small but idiomatic enhancement to the recursive explanation for Scala Hamming is ready for review.

ErikSchierboom (approved):

ErikSchierboom:
As you're saying this is more idiomatic, should this improved version be the version we show in the introduction/snippet/top of the approach?

bobahop:
Yes, I think you're right. Let me get rid of the substring stuff altogether...

bobahop:
@ErikSchierboom: Edits made for Hamming. Ready for review and hopefully merge. Still waiting on one other exercise to churn through Scala / Test / ci...

ErikSchierboom (approved):

PR #765, opened 2023-02-12
Food Chain approach

bobahop:
Approach for Food Chain


bobahop:
@ErikSchierboom: Scala Food Chain approach ready for review. Only like one approach, and I haven't seen it elsewhere.

ErikSchierboom (approved):
I'm fine with simple here! It's not as if the song will change lots :)

PR #790, opened 2023-07-12
Fix Yacht tests in online test runner

artamonovkirill:
Context: https://forum.exercism.org/t/scala-test-runner-reports-error-erroneously-for-prime-factors/6314/11

Test report is generated based on class name (=> `TEST-YachtTests.xml` in main), while the test runner expects `TEST-YachtTest.xml`.
The PR aligns class name with file name.

### DIFF
--- a/exercises/practice/yacht/src/test/scala/YachtTest.scala
+++ b/exercises/practice/yacht/src/test/scala/YachtTest.scala
@@ -1,7 +1,7 @@
 import org.scalatest.funsuite.AnyFunSuite
 import org.scalatest.matchers.should.Matchers
 
-class YachtTests extends AnyFunSuite with Matchers {
+class YachtTest extends AnyFunSuite with Matchers {
 
   test("Yacht") {
     Yacht.score(List(5, 5, 5, 5, 5), "yacht") shouldEqual 50


ErikSchierboom:
> Test report is generated based on class name (=> TEST-YachtTests.xml in main), while the test runner expects TEST-YachtTest.xml. 

As a follow-up, could we maybe look for a `*Test.xml` file to make it less error prone?

artamonovkirill:
> As a follow-up, could we maybe look for a `*Test.xml` file to make it less error prone?

That would probably help for "Prime Factors", but not for "Yacht" (where the class was named "YachtTest*s*").
I haven't dug too deep but it looks like Scala picks up whatever class (probably ones that extend `AnyFunSuite`).
Even
```scala
class YachtFooBar extends AnyFunSuite with Matchers {
  ...
}
``` is run and the result XML is `TEST-YachtFooBar.xml`.

I see two options:
- add a CI check for Scala project to make sure class name and file names match (also removes IDE warnings like below)
<img width="486" alt="Screenshot 2023-07-13 at 18 39 02" src="https://github.com/exercism/scala/assets/2599636/c63eea72-fa56-4058-a27b-15c9100b83d6">
- have the test runner read the test file (we already fetch the file name from config [here](https://github.com/artamonovkirill/scala-test-runner/blob/ee6ef9ea96520ffc8c863528a2d33a907b0e8968/bin/run.sh#L38C1-L45))

PR #802, opened 2024-01-25
Add reverse-string exercise

BNAndras:
@exercism/scala, this is related to the ongoing #48in24 year-long event Exercism is doing. This week features reverse-string.


ErikSchierboom (approved):
👍

PR #809, opened 2024-02-25
Add circular-buffer

BNAndras:
@ErikSchierboom, for #48in24. I adapted the Groovy implementation, and there's some visual noise in the track config because it hadn't been formatted previously.


ErikSchierboom (approved):
Brilliant!

PR #734, opened 2022-08-31
Scala3

LarsWestergren:
Some compiler warnings, should I fix those too?

Fixes https://github.com/exercism/scala/issues/684

PR #712, opened 2022-02-08
Bump Scala to version `2.13` and scalatest to version `3.2.10`

CJSmith-0141:
Addresses #678 and partially addresses #684 

I also ran `configlet sync -u`, not sure if that was needed or necessary. 

I was not able to get the generation code to work, could someone help bring that home? It might need a re-write. 

Help needed on the test generation. Manually editing the tests with these changes works.

PR #836, opened 2024-07-18
Scala 3 support

grzegorz-bielski:
TODO:
- [x] Fixing exercises
- [x] New test-gen POC
- [x] Scala 3 test runner: https://github.com/exercism/scala-test-runner/pull/55

Forum thread: https://forum.exercism.org/t/scala-3-support/12117

Supersedes: https://github.com/exercism/scala/pull/734


grzegorz-bielski:
I tried to port test generators to Scala 3, but looks like it's not going to be that easy.

I was able to update the libraries to the newest versions, but their up-to-date version simply doesn't work with the unsafe hack that was used in the test generators where `JsValue` from `play-json` was casted to `Map[String, Any]` . 
In order to make test generators working again they have to be refactored at the ground level and each individual test generator should be ported, one by one. 
It's not that hard, but requires a significant time commitment for which I'm not ready a the moment.
I already made a new POC setup based on the old one and ported one test generator for the `word-count` exercise.

My current development plan:
- for now I'll focus on porting only the `exercises` folder, just to make it compile with Scala 3. Even if something was auto-generated before, I won't touchits test-gen code. This way we can move forward, and people could start using Scala 3 in the exercism.
- I'll remove the old test-gen setup and create a new one, where we can port test-generators one by one.
From what I can see at: https://github.com/exercism/scala/pull/712 they were already not working 2 years ago, and recent new exercises are not even having test-gens.

@ErikSchierboom Does that sound good to you?

ErikSchierboom:
Looks good!

grzegorz-bielski:
@ErikSchierboom I fixed compiler errors in all exercises, please take a moment to review when you have a chance. 🙇 

There are still some warnings, but all tests are passing locally and I believe we can fix them later one by one when porting exercises to a more idiomatic Scala 3 and adding new test-gens.

I also noticed that current tests (bin/test) are _really_ slow. This is because we are starting a new `sbt` session for each exercise. We should consider refactoring it later. Perhaps something similar to [this](https://github.com/exercism/scala/blob/main/project/Build.scala), but for modern sbt. I removed it, since similarly to the test gens, it was already a dead code for some time now, but the idea though is sound.

ErikSchierboom (commented):
Great work!

Could you also update the documentation? In particular `docs/INSTALLATION.md`, but there might be other places that need to be updated too.

grzegorz-bielski:
@ErikSchierboom I updated the docs.

On the site there is also `Key Features of Scala`. I think it needs some updates too. The `Lazy computation` point is not really correct. We can take some inspiration from the official scala lang page: https://www.scala-lang.org/.
 Not sure where it can be changed, and what icons are available.

docs/ABOUT.md
```diff
@@ -1,9 +1,12 @@
 # About
 
-Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM and JavaScript runtimes let you build high-performance systems with easy access to huge ecosystems of libraries.
+A programming language that scales with you: from small scripts to large multiplatform applications.
+
+Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM, JavaScript and Native runtimes let you build high-performance systems with easy access to huge ecosystems of libraries.
 
 * Features
     * JVM Interoperability - Scala runs on the JVM, so Java and Scala stacks can be mixed for seamless integration.
+    * JS and Native interoperability - Thanks to the [Scala.js](https://www.scala-js.org/) and [Scala Native](https://github.com/scala-native/scala-native)
```
ErikSchierboom, line 9:
```suggestion
    * JS and Native interoperability - Scala can run in the browser and on native devides, thanks to [Scala.js](https://www.scala-js.org/) and [Scala Native](https://github.com/scala-native/scala-native)
```

docs/ABOUT.md
```diff
@@ -1,9 +1,12 @@
 # About
 
-Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM and JavaScript runtimes let you build high-performance systems with easy access to huge ecosystems of libraries.
+A programming language that scales with you: from small scripts to large multiplatform applications.
+
+Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM, JavaScript and Native runtimes let you build high-performance systems with easy access to huge ecosystems of libraries.
```
ErikSchierboom, line 5:
```suggestion
Scala is a a programming language that scales with you: from small scripts to large multiplatform applications.

Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM, JavaScript and Native runtimes let you build high-performance systems with easy access to huge ecosystems of libraries.
```

ErikSchierboom:
Here we go! 🥳

mrybak834:
Hey, I appreciate the update however this resulted in issues with implicit nulls, seen in exercise [two-fer](https://exercism.org/tracks/scala/exercises/two-fer/).

So I imagine there are other exercise issues related to this. Thanks!

Screenshot 2024-08-14 at 8 47 37 PM
Screenshot 2024-08-14 at 8 51 38 PM

mrybak834:
Thank you for this <3

ErikSchierboom:
> I also noticed that current tests (bin/test) are really slow. This is because we are starting a new sbt session for each exercise. We should consider refactoring it later. Perhaps something similar to [this](https://github.com/exercism/scala/blob/main/project/Build.scala), but for modern sbt. I removed it, since similarly to the test gens, it was already a dead code for some time now, but the idea though is sound.

Improving this would be great!

asarkar:
Hi, thanks for upgrading the Scala track. I started working on the exercises, and see that some of the tests are marked as `pending`. These tests are skipped locally, but occasionally fail when submitted, so, clearly the test runner doesn't care about the `pending` marker. Can you please advise how to run these tests locally so as not to have this discrepancy?

ErikSchierboom:
@asarkar That's documented here: https://exercism.org/docs/tracks/scala/tests

PR #866, opened 2025-07-15
ci uses ubuntu-24.04

keiravillekode:


keiravillekode:
Now using https://github.com/sbt/setup-sbt

.github/workflows/test.yml
```diff
@@ -8,11 +8,23 @@ on:
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+          cache: sbt
+
+      - name: Setup SBT
+        uses: sbt/setup-sbt@v1
+        with:
+          sbt-runner-version: 1.9.9
```
grzegorz-bielski, line 27: I don’t think we necessarily need to setup the sbt runner script version.
Otherwise LGTM.

grzegorz-bielski (approved):
Thanks for the help!

PR #874, opened 2026-04-15
bin/verify-exercises-in-docker

keiravillekode:
This is a useful script for testing exercises against a test runner.


rabestro (approved):
Thanks for the script!

